PR #9, opened 2019-07-25
Simplify the content API and add missing type support

MangelMaxime:
Children can be a string, bool, int, ReactElement, ReactElement list, etc.

This avoids an unnecessary wrapping

```ts
    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```


Zaid-Ajaj:
Thanks @MangelMaxime :fire:

PR #28, opened 2019-08-24
Use createObj instead of keyValueList

### LINKED ISSUE #17: Incorrect use of "keyValueList CaseRules.LowerFirst"?
AFAIK `keyValueList CaseRules.LowerFirst` is only for transforming a DU list into an object. Since `IReactProperty` and `IStyleAttribute` are just `string * obj` in disguise, `createObj` is probably the right choice in Feliz.

Please tell me if this is not right, because then I have misunderstood something.

cmeeren:
Fixes #17

Untested

### DIFF
--- a/demo/App.fs
+++ b/demo/App.fs
@@ -55,7 +55,7 @@ let update msg state =
     style.color.red
     style.color "#000000"
 ]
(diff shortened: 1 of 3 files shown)


Zaid-Ajaj:
Thank you, I will test it myself

PR #29, opened 2019-08-24
Add sizes and srcset props

### LINKED ISSUE #20: Missing sizes and srcSet props
There doesn't seem to be any `sizes` or `srcSet` props (for the `img` element).

cmeeren:
Fixes #20 

Just thin string wrappers, because there are so many variants of the syntax, particularly for `sizes`.


Zaid-Ajaj:
Fixed in [this commit](https://github.com/Zaid-Ajaj/Feliz/commit/e13d597b0d2752579e820954d5e9765c51cf3928) though for now only the string wrappers

PR #73, opened 2019-10-27
Remove empty dependency parameter from default useEffect, fixes #72

### LINKED ISSUE #72: useEffect is only firing on initial render but not on later updates
`useEffect` is not firing on every update. I believe it is supposed to by default -- [React docs here](https://reactjs.org/docs/hooks-effect.html).

Here is repro code -- the Fable.React version fires on every increment rerender, but the Feliz version only fires on the initial render:

```fs
let counter =
    React.functionComponent("Counter", fun () ->
        let (count, setCount) = React.useState(0)

        // prints on every increment
        Fable.React.HookBindings.Hooks.useEffect ( fun _ -> printfn "useEffect from Fable.React" )

        // prints only on initial render
        Feliz.React.useEffect ( fun _ -> printfn "useEffect from Feliz.React" )

        Html.div [
            Html.h1 count
            Html.button [
                prop.text "Increment"
                prop.onClick (fun _ -> setCount(count + 1))
            ]
        ]
    )
```

zanaptak:


Zaid-Ajaj:
Awesome, thanks a lot @zanaptak

PR #74, opened 2019-10-28
Updated comments and added height enums

Shmew:


Zaid-Ajaj:
Great stuff @Shmew, merged and published as of v0.65.3 :rocket:

PR #76, opened 2019-10-28
Feliz.ElmishComponents: Initial Release

Zaid-Ajaj:
This little gem will be really useful from here on now :heart: let us experiment 🚀


Zaid-Ajaj:
Docs here: https://zaid-ajaj.github.io/Feliz/#/Feliz/ElmishComponents

cmeeren:
I don't get it. The docs seems to focus on simplifying "child component" stuff, but the sample doesn't show any child components.

Zaid-Ajaj:
I need to write a bigger example but the idea is that you have an internal dispatch loop so that you don't need to clutter the application structure, something like this
```fs
let render state dispatch = 
  match state.CurrentUrl with 
  | [ "login" ] -> Login.login(signedIn = fun user -> dispatch (UserSignedIn user))
  | [ "home" ] -> Home.home(user = state.User)
  | // etc.
```
where the parent component that is (in the example above) keeping track of the `CurrentUrl` and `User` doesn't need to know about implementation details from `Login.login` nor `Home.home` (the child components) 

This is not the best example of course because I am missing the part in the `update` function where I programmatically navigate to `login` when there is not user in this parent component.

Zaid-Ajaj:
Sorry if this is vague, I thought it would be "obvious" once you know that you can build standalone components that maintain their own state in their own dispatch loop without having the parent components manage them explicitly. I think writing a component in Elmish style and connecting it to the parent using props (Common React pattern) is the combination from the best of both worlds. Of course I need to experiment with it more to understand the limitations and how to make it even better and nicer to work with

cmeeren:
What I'm very interested in is how to manage state that is needed by multiple "components".

I am currently (still) of the opinion that splitting an Elmish app into fully separate MVU components is not to be taken lightly. It's not *necessarily* a bad thing, but it's something that should only be done very deliberately. I have based this on (some) personal experience as well as [these](https://www.reddit.com/r/elm/comments/5xdl9z/elm_architecture_with_a_reduxlike_store_pattern/dehrcx8/) two [answers](https://www.reddit.com/r/elm/comments/5jd2xn/how_to_structure_elm_with_multiple_models/dbuu0m4/) by rtfeldman. In short, if a "child component" needs state from its parent, that state must be duplicated and updated in the child component, too. Furthermore, a child component may need to communicate with its parent. Usually, in MVU, a better solution is to split the model/message/update/view separately and only when required. I explain this briefly in the Elmish.WPF tutorial section [Child components and scaling](https://github.com/elmish/Elmish.WPF/blob/master/TUTORIAL.md#child-components-and-scaling).

If ElmishComponents can alleviate this, I'd be very interested to know how. And again, I'm sorry if I have misunderstood this library.

Shmew:
I thought the point of this was that when those cases don't apply, meaning they don't need to communicate with the parent is where this is beneficial. Wouldn't we be able to just pass the parent state into the component if it needs it? 

I know that whenever I have children Elmish stuff I always use lazyOf2..3..etc is this needed/applicable with the new library?

Zaid-Ajaj:
> if a "child component" needs state from its parent, that state must be duplicated and updated in the child component, too. Furthermore, a child component may need to communicate with its parent.

> If ElmishComponents can alleviate this, I'd be very interested to know how. And again, I'm sorry if I have misunderstood this library.

@cmeeren 

That is exactly what ElmishComponents tries to solve: avoids duplication altogether. I made a bigger example, implemented twice to showcase the implementation difference between traditional elmish (with state duplication) and "hybrid" Elmish with React components and how ElmishComponents can simplify communication between parent and children using React patterns, see [elmish-composition](https://github.com/Zaid-Ajaj/elmish-composition) repository that now contains both applications! 

Please take a look at the examples there and let me know what you think! 

In summary the big difference is in this `App` parent component that acts as glue for it's children, the component itself doesn't do a lot, it only switched from `Login`to `Home` once a login attempt is succesful:

- [Traditional App.fs](https://github.com/Zaid-Ajaj/elmish-composition/blob/master/traditional/src/App.fs) 

- [Hybrid App.fs](https://github.com/Zaid-Ajaj/elmish-composition/blob/master/hybrid/src/App.fs)

>  know that whenever I have children Elmish stuff I always use lazyOf2..3..etc is this needed/applicable with the new library?

@Shmew 

I am not sure I have seen anyone using `lazyOf` in application code (how does that look like?)

Shmew:
Sorry it's `lazyView` not `lazyOf` [here is an example](https://github.com/Shmew/MordhauBuddy/blob/master/src/Electron/Renderer/App/View.fs#L110).

Zaid-Ajaj:
Ah I see. `lazyView` is just memoization for your view. Using Feliz.ElmishComponents you can remove the submodels from the main model because the child components will be handling their data internally, see the example I commented above :smile:

cmeeren:
I just had a brief look at the elmish-composition repo. I will need to investigate and experiment further, but here are some immediate (and not too carefully formulated) thoughts:

* The "traditional" approach doesn't seem to contain duplication in the way I mean. Having `Login.Model` and `Home.Model` in the `App.Model` isn't duplication; it's composition. All state lives in only one place. (Okay, `CurrentUser` in `App` duplicates state from `Home`, but it's not used for anything and can be removed.) I am talking about a "child component" (e.g. a book detail view) needing access to state from a "parent component" (e.g. a all books from the parent book list view in order to do auto-completion of authors etc.). Then, the list of books must live both in the "parent" and the "child" component, and updates to the list (which happen in the parent) must be propagated to the child component, too, using a dedicated child message which sets the child component's list. It gets painful right off the bat.

* The "hybrid" example allows illegal state (`Page.Home` without a user), though that isn't necessary. Using a proper DU (not StringEnum) for `Page` with `Page.Home of User`, and removing the `User` from the model, makes it function the same but makes the illegal state impossible.

* The "hybrid" approach seems to not really do anything you couldn't do by scaling M, V, and U separately and passing whatever is needed. But I need to experiment a bit more to get to the core of the parallels here.

* I don't really like passing a "logout" function as a prop in the "hybrid" approach. I can't put my finger on exactly why "dependency injection" like that seems to against the grains as it were, but my gut tells me it'll have drawbacks at scale (plus, DI makes things harder to test than pure data).

* My main point, though: I don't really know what you mean by "traditional". The example is factored using a fully separate M, V, and U, i.e. a "child component", which AFAIK is not the recommended way to scale Elmish apps (see my two previous links). The "traditional" app could be a lot nicer if factored according to those tips. As it stands (at least as far as a comparison base for "hybrid"), it's essentially a strawman. I'd love to have a go at trying to factor this according to the previously linked recommendations, but I have no idea when I'll be able to get to that. In the meantime I encourage others to try the same and share their findings.

I don't want to drive my points too hard, because as I said, I need to investigate this further and experiment for myself.

Zaid-Ajaj:
> The "traditional" approach doesn't seem to contain duplication [...]

Yes, you are right, this is not a duplication example but rather a plain old (i.e. traditional) program composition. 

> The "hybrid" example allows illegal state (Page.Home without a user), though that isn't necessary. Using a proper DU (not StringEnum) for Page with Page.Home of User, and removing the User from the model, makes it function the same but makes the illegal state impossible.

Now fixed, doesn't have anything to do with the approach itself, just slacking from my side, should be a proper implementation now with changes from [this commit](https://github.com/Zaid-Ajaj/elmish-composition/commit/e492bf9a1ff3564f179c332666d0e1072d48d629) please check it out! 

> The "hybrid" approach seems to not really do anything you couldn't do by scaling M, V, and U separately and passing whatever is needed.

I think *this* is the confusion point here: elmish-components doesn't will now allow you to do extra magic that is otherwise impossible with scaling M, V or U. Instead, it makes the composition process *a lot* simpler but not having the parents track and manage states from their children. In other words, it allows you to build React components (which have internal state) using TEA.  

Now the parent `App` only does the switching between the components without the boiler-plate: so damn simple!
```fs
type Page =
    | Login
    | Home of SignedInUser

type State = { CurrentPage : Page }

type Msg =
    | SignedIn of SignedInUser
    | Logout

let init() =
    let initialState = { CurrentPage = Login }
    initialState, Cmd.none

let update (msg: Msg) (state: State) =
    match msg with
    | SignedIn user -> { state with CurrentPage = Home user  }, Cmd.none
    | Logout -> init()

let render (state: State) (dispatch: Msg -> unit) =
    match state.CurrentPage with
    | Login -> Login.login { signedIn = SignedIn >> dispatch }
    | Home user -> Home.home { user = user; logout = fun _ -> dispatch Logout  }
```

> I don't really like passing a "logout" function as a prop in the "hybrid" approach. I can't put my finger on exactly why "dependency injection" like that seems to against the grains as it were, but my gut tells me it'll have drawbacks at scale (plus, DI makes things harder to test than pure data).

This is a very common React pattern, passing event through the props of component to communicate things that have happened and letting the parent component "react" upon them. I don't know why this would the `update` function *less* unit testable though

> I'd love to have a go at trying to factor this according to the previously linked recommendations, but I have no idea when I'll be able to get to that. In the meantime I encourage others to try the same and share their findings.

Please do when you have the time! I know this example can be simplified but this is the core of how composition works in Elm, whether you simplify the model the parent and children into one thing, the explicit passing of everything, the mapping of the commands from child to parent etc. 

To summarize the punch-line of ElmishComponents: 
 - Allows for build React components following The Elm Architecture
 - Simplifies boilerplate in program composition

cmeeren:
> I think this is the confusion point here: elmish-components doesn't will now allow you to do extra magic that is otherwise impossible with scaling M, V or U. Instead, it makes the composition process a lot simpler but **not having the parents track and manage states from their children**. In other words, it allows you to build React components (which have internal state) using TEA.

(Emphasis mine) Yes, I think this is the confusion. When you say "having the parents track and manage child state" you seem to be talking about a completely separate MVU component with its own model/msg/view/update. This is the kind of scaling that's very easy to reach for, but which all the links I have posted says to steer away from because it's counterproductive. The scaling I am talking about does not split all three of M, V, and U at the same time. It splits only those that become so large they are painful to work with, one at a time, as described [here](https://www.reddit.com/r/elm/comments/5jd2xn/how_to_structure_elm_with_multiple_models/dbuu0m4/).

> I don't know why this would the `update` function _less_ unit testable though

To be clear, I'm not saying it's *not* testable. But testing pure data input/output is simpler than having to mock functions to return something or verify that some input has been received (if nothing else, the unit tests become more verbose).

Zaid-Ajaj:
> This is the kind of scaling that feels very natural and easy to reach for, but which all the links I have posted says to steer away from because it's counterproductive in FP/MVU.

This is the type of scaling I am trying to simplify using known patterns from React and Vue.js: removing the "counter productive" part of that statement. 

> The scaling I am talking about does not split all three of M, V, and U at the same time. It splits only those that become so large they are painful to work with, one at a time, and only when necessary, as described here.

Yes, that is the better way to Elm composition but I think *there is* value to `Feliz.ElmishComponent` for those who think in separate components, which is what the rest of the JS is world is doing and not necessarily writing pure 100% MVU architecture

PR #78, opened 2019-10-28
Add transitionProperties and transitionDelay overloads

zanaptak:
Added remaining properties from [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties) -- except for the vendor prefixed ones (`-moz`, `-webkit`), but if you want, I can add them too.

Also cleaned up some spelling, whitespace, and sorting.

Added the `transitionDelay` overloads to match `transitionDuration`.


Zaid-Ajaj:
Awesome stuff @zanaptak, patched and published Feliz as of version 0.65.4 :rocket: should be available soon on nuget. 

> except for the vendor prefixed ones (-moz, -webkit), but if you want, I can add them too.

I am not sure if we want to add these because it is not just this property that requires them but also a lot more styles so let's keep it like this for a while till someone asks for it ;)

PR #82, opened 2019-10-29
WIP Add React.useCallback, React.useMemo, React.memo

zanaptak:
`useCallback` and `useMemo` are fairly straightforward.

`React.memo` deserves some discussion.

Currently this essentially matches the [React API](https://reactjs.org/docs/react-api.html#reactmemo) (`React.memo(render, areEqual)`) using optional params: `React.memo(render, ?areEqual, ?displayName)`. (Display name is exclusive to Fable, to provide debug component names for function values.)

It works either for wrapping a React.functionComponent, or as a complete replacement:

```fs
// Wrapping
let counter =
    React.functionComponent("Counter", fun props ->
        Html.div [
            // ...
        ]
    )
let memoWrapCounter = React.memo(counter, displayName="memoWrapCounter")

// Replacement
let memoCounter =
    React.memo(fun props ->
        Html.div [
            // ...
        ]
    , areEqual=myCustomEqualityFn // optional
    , displayName="memoCounter" // optional
    )
```

By default (omitting the `areEqual` param) it is just letting React use its own default internal equality compare. I'm not sure if the `(memo)EqualsButFunctions` functions from Fable.React are very necessary by default for this library. Possibly users can instead use `useCallback` to create stable function references that won't break default equality. In a quick test it seemed to work fine for composed Elmish dispatchers e.g. `{ props.childDispatch = React.useCallback((ChildMsg >> dispatch), [||]) }`.

This current argument order however is inconsistent with the existing `React.functionComponent` which has an overload for display name first. So probably a maintainer decision regarding internal consistency vs. API conformity.

There is also the question whether to keep it as a separate `React.memo`, or be more like the Fable.React approach and have it combined with the `React.functionComponent` using additional overloads and/or optional params.

### DIFF
--- a/Feliz/React.fs
+++ b/Feliz/React.fs
@@ -8,6 +8,44 @@ module internal ReactInterop =
     let useEffectWithDeps (effect:  obj) (deps: obj) : unit = import "useEffectWithDeps" "./ReactInterop.js"
     let useEffect (effect: obj) : unit =  import "useEffect" "./ReactInterop.js"
 
+type internal Internal() =
+    static let propsWithKey (withKey: ('props -> string) option) props =
+        match withKey with
+        | Some f ->
+            props?key <- f props
+            props
+        | None -> props
+    static member
+        functionComponent
+        (
+            renderElement: 'props -> Fable.React.ReactElement,
+            ?name: string,
+            ?withKey: 'props -> string
+        )
+        : Fable.React.FunctionComponent<'props> =
+            let functionElementType = Fable.React.ReactElementType.ofFunction renderElement
+            name |> Option.iter (fun name -> functionElementType?displayName <- name)
+            fun props ->
+                let props = props |> propsWithKey withKey
+                Fable.React.ReactElementType.create functionElementType props []
+    static member
+        memo
+        (
+            renderElement: 'props -> Fable.React.ReactElement,
+            ?name: string,
+            ?areEqual: 'props -> 'props -> bool,
+            ?withKey: 'props -> string
+        )
+        : Fable.React.FunctionComponent<'props> =
+            let memoElementType =
+                match areEqual with
+                | Some areEqual -> Fable.React.ReactElementType.memoWith areEqual renderElement
+                | None -> Fable.React.ReactElementType.memo renderElement
+            name |> Option.iter (fun name -> memoElementType?displayName <- name)
+            fun props ->
+                let props = props |> propsWithKey withKey
+                Fable.React.ReactElementType.create memoElementType props []
+
 type React =
     /// The `useState` hook that create a state variable for React function components.
     static member useState<'t>(initial: 't) = Interop.reactApi.useState(initial)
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj:
Hello @zanaptak Thanks a lot for taking on this issue which gets up closer to the stable release :rocket: 

> useCallback and useMemo are fairly straightforward.

True, however it would be nice to have overloads that make the dependency array optional, defaulting to an empty array what do you think? 

As for `React.memo`, the general guide line is to stay as close as possible to the React equivalent which what you are doing 💯 as for `displayName` I would rather have it as first argument and the render implementation the last argument.

> Possibly users can instead use useCallback to create stable function references that won't break default equality. In a quick test it seemed to work fine for composed Elmish dispatchers e.g. { props.childDispatch = React.useCallback((ChildMsg >> dispatch), [||]) }.

We will need to document this properly: when and how to use these functions. Ideally with examples. Is that something you maybe can help me with? Would be really awesome! 

> There is also the question whether to keep it as a separate React.memo, or be more like the Fable.React approach and have it combined with the React.functionComponent using additional overloads and/or optional params.

I would say keep it as a separate `React.memo` to follow Reacts patterns

zanaptak:
I can make these changes. With render last, do you want all combinations like so?
```fs
React.memo(name, areEqual, render)
React.memo(name, render)
React.memo(areEqual, render)
React.memo(render)
```

Also do you want the `withKey` parameter like Fable.React? It's not strictly necessary since key can be provided in props, but arguments for it are at https://github.com/fable-compiler/fable-react/issues/167. Not sure I'm convinced personally but others might expect/want it for flexibility.

What do you think about only bringing `name` forward (as probably most commonly used), to consolidate the overloads a bit?
```fs
React.memo(name, render, ?areEqual, ?withKey)
React.memo(render, ?areEqual, ?withKey)
```

Shmew:
IMO I think additional overloads as opposed to optional parameters are nicer since the resulting code is less messy and it doesn't change bundle sizes.

zanaptak:
I find the opposite. Args as Options are clean and natural in F#. The F# style guide prefers optional args before overloads. There is less duplication of signatures and XML docs. But it is subjective and I will of course defer to whatever decision is made.

Shmew:
Yeah, I mean I never use overloads normally, but when the entire library is built with them as a feature it feels cleaner and more consistent.

Zaid-Ajaj:
> I can make these changes. With render last, do you want all combinations like so?

Yes this looks good, my reason for this is that the render function is usually multi-line, so to put the end you get code like this:
```fs
React.memo("App", fun props ->
  Html.div [
    Html.h1 "Hello there" 
  ])
```
Another overload for `render` of type `props -> ReactElement list` would also be nice so that users don't have to wrap their components by dummy `divs`:
```fs
React.memo("App", fun props -> [
    Html.h1 "Hello there" 
    Html.br [ ]
    Html.p "More content here"
])
```
this overload would call the first overload that takes `prop -> ReactElement` and wraps the list into `Html.fragment` 

> Also do you want the withKey parameter like Fable.React? It's not strictly necessary since key can be provided in props, but arguments for it are at fable-compiler/fable-react#167. Not sure I'm convinced personally but others might expect/want it for flexibility.

Hmm I am not sure personally but I guess it is fine as long as we document these features properly so that users know what to expect.

@Shmew Aren't optional parameters a short-hand for overloads? Again no preference for me here because the surface API stays the same and we will document it anyways

Shmew:
@Zaid-Ajaj  Yeah they basically are, I'm not at all opposed to optional parameters. I prefer them when I write my own code. The `React` methods are different from how the `Html` stuff is designed anyways so it probably is best to just use optional parameters.

zanaptak:
Required parameter in last position means we can't use optional args anyway so I added the billion overloads.

I added the `withKey` and the `props -> seq` overloads to `functionComponent` too.

There is a small issue with the `seq` overloads: the left pipe syntax no longer works. The following fails to compile with error `A unique overload for method 'functionComponent' could not be determined based on type information prior to this program point. A type annotation may be needed.`:
```fs
let greeting = React.functionComponent <| fun (props: Greeting) ->
    Html.div [
        Html.span "Hello, "
        Html.span (Option.defaultValue "World" props.Name)
    ]
```

You have to parenthesize to get the type inference to work:
```fs
let greeting = React.functionComponent(fun (props: Greeting) ->
    Html.div [
        Html.span "Hello, "
        Html.span (Option.defaultValue "World" props.Name)
    ])
```

Zaid-Ajaj:
> There is a small issue with the seq overloads: the left pipe syntax no longer works. The following fails to compile with error A unique overload for method 'functionComponent' could not be determined based on type information prior to this program point. A type annotation may be needed.:

It is not really an issue but it is a small breaking change which is why the docs application isn't building anymore. That is fine, I will fix it myself, but for now, let's get this goodness merged in :smile:

PR #85, opened 2019-10-31
Add hsl to colors

Shmew:


Feliz/Colors.fs
```diff
@@ -2,6 +2,8 @@ namespace Feliz
 
 /// Contains a list of HTML5 colors from https://htmlcolorcodes.com/color-names/
 module colors =
+    let hsl (hue: float, saturation: float, lightness: float) =
```
cmeeren: I recommend renaming `saturation` and `lightness` to `saturationPercent` and `lightnessPercent`, to make it clear that they should be in the interval [0, 100]. Perhaps also add a doc comment mentioning this.

PR #103, opened 2019-11-03
Rename from SingleAngle to StraightAngle

Martin4ndersen:
Rename SingleAngle example to StraightAngle as it's called this in the [Recharts example](http://recharts.org/en-US/examples/StraightAnglePieChart).


Zaid-Ajaj:
Awesome! thanks a lot @Martin4ndersen I have merged and published it to the docs, the changes should be live in any minute now :smile:

PR #113, opened 2019-11-13
Fix border styles

Shmew:
All css properties taking IBorderStyle are currently broken since this sets the border styles to IStyleAttribute.

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -1464,55 +1464,55 @@ module style =
         /// Specifies a dotted border.
         ///
         /// See example https://www.w3schools.com/cssref/playit.asp?filename=playcss_border-style&preval=dotted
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
They are not broken, they are supposed to be used as such and return `IStyleAttribute`:
```fsharp
Html.div [
  prop.style [ style.borderStyle.dashed ]
]
```
> This is why the build is failing, the docs app uses this

But sometimes you have other style properties which have `IBorderStyle` as input such as `style.border` and then you can use `IBorderStyle` as one of the input parameters:
```fsharp
Html.div [
  prop.style [ style.border(3, borderStyle.dashed, color.crimson) ]
]
```
Here, `style.border` takes the second parameter of type `IBorderStyle`

Shmew:
Oh, I see now. Whoops.

PR #117, opened 2019-11-14
Add Html target attribute 

Shmew:


Feliz/Properties.fs
```diff
@@ -1137,6 +1137,19 @@ module prop =
...
+        /// Opens the linked document in a new window or tab.
+        static member inline blank = Interop.mkAttr "target" "_blank"
+        /// Opens the linked document in the same frame as it was clicked (this is default).
+        static member inline self = Interop.mkAttr "target" "_self"
+        /// Opens the linked document in the parent frame.
+        static member inline parent = Interop.mkAttr "target" "_parent"
+        /// Opens the linked document in the full body of the window.
+        static member inline top = Interop.mkAttr "target" "_top"
+        /// Opens the linked document in a named frame.
+        static member inline frameName (value: string) = Interop.mkAttr "target" value
```
cmeeren: This should be just a normal prop in the `prop` type, taking a `frameName: string` parameter.

Shmew: Wouldn't that make it more ambiguous? They'd be doing `prop.target.prop "someFrameName"` vs `prop.target.frameName "someFrameName"`

cmeeren: No, I'm talking about this:

```f#
prop.target "someFrameName"  // normal prop
prop.target.blank  // enum prop
```

Shmew: Oh, of course! Yes, that absolutely is more correct. Thank you I'll fix that.

Zaid-Ajaj:
Thanks @Shmew :heart:

PR #120, opened 2019-11-16
Add prop.step and min/max float overloads

zanaptak:
Step is used for range input type. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range


Zaid-Ajaj:
Thanks a lot @zanaptak

PR #122, opened 2019-11-18
Fixing typo in Nav element

Dzoukr:
I believe that Nav should be generated as `nav` instead of `label`.


Zaid-Ajaj:
Nice catch! thanks Roman 🎉

PR #123, opened 2019-11-19
ICssUnit overloads for prop.height and prop.width

Shmew:


Feliz/Properties.fs
```diff
@@ -365,10 +365,18 @@ type prop =
     ///
     /// <canvas>, <embed>, <iframe>, <img>, <input>, <object>, <video>
     static member inline height(value: int) = Interop.mkAttr "height" value
+    /// Specifies the height of elements listed here. For all other elements, use the CSS height property.
+    ///
+    /// <canvas>, <embed>, <iframe>, <img>, <input>, <object>, <video>
```
cmeeren: I think you have to add backticks around these tags, otherwise they won't display properly. e.g.

```f#
/// `<canvas>`, `<embed>`, ...
```

I see this is an issue with the existing overloads, too.

Zaid-Ajaj:
Hello @Shmew, I deliberately haven't added `ICssUnit` to those properties because these global HTML attributes do not accept anything other than, and I quote the [HTML5 specs](https://html.spec.whatwg.org/#attr-dim-height), **valid non-negative integers**

> Author requirements: The width and height attributes on img, iframe, embed, object, video, and, when their type attribute is in the Image Button state, input elements may be specified to give the dimensions of the visual content of the element (the width and height respectively, relative to the nominal direction of the output medium), in CSS pixels. The attributes, if specified, must have values that are **valid non-negative integers**.

Shmew:
Oh, I didn't realize that changed with html5. I'll start just opening issues before I make PRs 😅

PR #130, opened 2019-12-03
Add autoCapitalize prop

Shmew:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

### DIFF
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -1408,6 +1408,7 @@ module prop =
         /// applied.
         static member inline other = Interop.mkAttr "aria-sort" "other"
 
+    [<Erase>]
     type type' =
         /// Defines a password field
         static member inline password = Interop.mkAttr "type" "password"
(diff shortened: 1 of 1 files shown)


Feliz/Properties.fs
```diff
@@ -1453,3 +1454,14 @@ module prop =
         static member inline url = Interop.mkAttr "type" "url"
         /// Defines a week and year control (no timezone)
         static member inline week = Interop.mkAttr "type" "week"
+
+    [<Erase>]
+    type autoCapitalize =
```
cmeeren: Should be lowercase: `autocapitalize`

Feliz/Properties.fs
```diff
@@ -1453,3 +1454,14 @@ module prop =
...
+        /// No autocapitalization is applied (all letters default to lowercase)
+        static member inline off = Interop.mkAttr "autoCapitalize" "off"
```
cmeeren: All of these should be lowercase: `autocapitalize`

Shmew: I thought so too at first, but then I saw how other props in this file use the same casing and it works as expected. Such as [this property](https://github.com/Shmew/Feliz/blob/patch-1/Feliz/Properties.fs#L349). I tested it out and it does apply the attribute in all lower case so I figured I should follow the style of the rest of the file.

cmeeren: Oh, right. As long as it works :)

PR #132, opened 2019-12-10
Adds prop.none as counterpart of Html.none

Dzoukr:
Can be hand when generating props based on some model (sometimes you don't need yield any, e.g. when validation is ok)

I am not sure if needs to be `inline` though.

### DIFF
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -40,6 +40,8 @@ type AriaRelevant =
 
 /// Represents the native Html properties.
 type prop =
+    /// The empty property, adds no attribute to element.
+    static member inline none = Interop.mkAttr "" ""
     /// Often used with CSS to style a specific element. The value of this attribute must be unique.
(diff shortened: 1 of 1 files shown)


Dzoukr:
Oh, now I see it throws `Warning: Invalid attribute name: '' ` in react. Maybe to support this, there must be some filtering over attribute name is not empty?

Zaid-Ajaj:
Why do you need this? You can just write 
```fs
Bulma.button [
  if state.Flag then prop.something
  prop.somethingElse
  // etc.
]
```
If you are moving the props to a function, you can write this
```fs
let buttonProps state = 
  if state.Flag 
  then [ prop.something ]
  else [ ] 

Bulma.button [
    yield! buttonProps state
    yield prop.somethingElse
] 
```

Dzoukr:
Yeah, I can do it using `yield!`, but since there is `Html.none`, I was just thinking about having counterpart for props. But can surely work with `yield!` instead, no problem.

PR #139, opened 2020-01-22
Clarify prop vs. element list

### LINKED ISSUE #138: SUGGESTION: Make the use of `prop.children` explicit when using other `prop.<X>`
I think it would make sense to include in the docs (don't know where exactly though), an _explicit_ mention that once a user puts _any_ `prop.<attribute>` inside a tag, if other tags are required, not just more attributes, it **must** be done via `prop.children []`.  
(This is true since the `prop.<X>` type is `IReactProperty` whereas tags are `ReactElements`, leading to type mismatch conflicts).

I hit this issue, and, although I had the docs from the official site right in front of me, it took me a while to figure it out.

Something on the lines of:
```fs
Html.div
    [ prop.classes ["MyClassName"]
      Html.div [...] ] // --> type mismatch with the `prop.classess` element
```
won't compile due to type mismatch. You need to do it like:
```fs
Html.div
    [ prop.classes ["MyClassName"]
      prop.children
           [ Html.div [...] ] ]
```
which will.

cmeeren:
Fixes #138

### DIFF
--- a/public/Feliz/Syntax.md
+++ b/public/Feliz/Syntax.md
@@ -24,7 +24,9 @@ Html.h1 [
     ]
 ]
 ```
-Which is the same as:
+Note that you can't mix attributes and child elements. If you use the component overload taking `IReactProperty list`, you must use `prop.children` to specify the child elements, as shown above.
+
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Thanks a lot @cmeeren

PR #145, opened 2020-02-26
Remove height entry in pigeonmaps.fs

Gilotto:
While working with a pigeonMaps map in a div, I wanted to size the map according to parent div. (which is based on length.vh(80))

Desired behaviour: resize map according to parent div

Current behaviour: Map sizes to 200px height (the default in PigeonMaps.fs)

The default height property of 200 interferes with the inherited height of the parent div, see image below.

afbeelding

### DIFF
--- a/Feliz.PigeonMaps/PigeonMaps.fs
+++ b/Feliz.PigeonMaps/PigeonMaps.fs
@@ -18,7 +18,7 @@ type PigeonMaps =
         let defaults = createObj [
             "center" ==> [| 40.416775; -3.703790 |]
             "zoom" ==> 3
-            "height" ==> 200
+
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hello @Gilotto, thanks a lot for the fix! I was wondering how we could set a default height of the div if someone didn't specify anything, but I guess it is expected behavior that a div should always take the height of it's parent elements when it is not specified on the div itself

Gilotto:
Hey @Zaid-Ajaj , Thanks for your quick answer and merge!

The intellisense for the height property (style.height) mentions the inheritence.
It is working as intended now.

afbeelding

Issue #81, opened 2019-10-28
Alphabetize props/styles/etc


Shmew:
A lot of the entries aren't in a defined order, having them alphabetized would make maintenance easier going forward.

Zaid-Ajaj:
This is surely be nice to have! Although it is low-priority, would be awesome if someone wants to take it :smile:

PR #150, opened 2020-03-16
Complete HTML elements and attributes

### LINKED ISSUE #124: Missing iFrame attributes
Many of these properties are currently not present:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

### LINKED ISSUE #149: Missing support for Html.Source and Video props
First of all, thank you for such awesome library, which in contrast to the default Elmish components is really friendly to use with the autocomplete; right now I am using it to create a personal blog and the experience has been great so far.

I see that most of the attributes for the video html tag are missing as well as the whole Source html element. Video and Source element specification:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

Right now the easiest workaround seems to use `prop.dangerouslySetInnerHTML` with string templating as I do not see the way to set a custom made Html tag and set Html properties which are not a string key-pair.

``` F#
prop.dangerouslySetInnerHTML (sprintf """
<video poster="%s" loop autoplay muted >
      <source src="videos/IMG_5202.mp4" type="video/mp4">
</video>
""" backgroundPath)
```

My guess is that here could be the implementation together with the rest of the tags, but as I am still new to F# and I have never ever touched the react API before I am not that confident to make a PR to
`master/Feliz/Html.fs`

Shmew:
* Added all HTML 5.2 spec elements and their respective attributes.
* Organized and made uniform code structure for Properties and Html.
* Added extra overloads where only `int` was specified and applicable.
* Added some additional comments outside the functions and types defined where they were missing.

Should resolve #149 and resolve #124
Partial work towards #81

### DIFF
--- a/Feliz/BorderStyle.fs
+++ b/Feliz/BorderStyle.fs
@@ -5,18 +5,14 @@ open Feliz.Styles
 
 [<Erase>]
 type borderStyle =
-    /// Specifies a dotted border.
-    ///
-    /// See example https://www.w3schools.com/cssref/playit.asp?filename=playcss_border-style&preval=dotted
-    static member inline dotted : IBorderStyle = unbox "dotted"
     /// Specifies a dashed border.
     ///
     /// See example https://www.w3schools.com/cssref/playit.asp?filename=playcss_border-style&preval=dotted
     static member inline dashed : IBorderStyle = unbox "dashed"
-    /// Specifies a solid border.
+    /// Specifies a dotted border.
     ///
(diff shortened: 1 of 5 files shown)


Zaid-Ajaj:
Hello @Shmew, this PR looks really good, I would like to add it in ASAP. 

Would you please try to rebase your branch on master to fix the merge conflicts?

Zaid-Ajaj:
Awesome stuff, thanks a lot @Shmew, I will publish ASAP :rocket: :fire:

PR #152, opened 2020-03-25
Add missed html props

Shmew:
Added a few missed props.


Zaid-Ajaj:
Thank you :pray: will publish a nuget later this evening :smile:

Issue #153, opened 2020-03-29
Maxlength causes DOM error.


DurdSoft:
I believe `maxLength` is named wrong here:

https://github.com/Zaid-Ajaj/Feliz/blob/440d32527c63ab3ef2b10a9922df4998b38517fb/Feliz/Properties.fs#L578

It is `"maxlength"` and should be `"maxLength"`. This then causes a DOM error:

![image](https://user-images.githubusercontent.com/24255513/77847683-e351d900-71b6-11ea-807f-770193c9dd08.png)

PR #154, opened 2020-03-29
maxlength -> maxLength

DurdSoft:
I believe this will fix https://github.com/Zaid-Ajaj/Feliz/issues/153


Zaid-Ajaj:
Thanks a lot @DurdSoft I have released your fix as of v0.74 and it will available soon :smile:

PR #161, opened 2020-04-14
Add Recharts bindings for optional data values and axis scale type

joahim:
This pull request adds Recharts binding for:

- Optional data values (e.g. https://recharts.org/en-US/examples/LineChartConnectNulls)
- Setting the x/y axis scale type (https://recharts.org/en-US/api/XAxis#scale)

Optional data values examples added to the docs for LineChart and AreaChart.


Zaid-Ajaj:
Hello @joahim, this is really awesome work, thanks a lot :heart: 

I only have one comment about the `Scale` type, it is defined as a separate type to be used as input for the `xAxis.scale` function so you would use like this: `xAxis.scale Scale.Linear`. I would rather have it defined the same way  the `orientation` is defined with static properties to make easily it discoverable and be able to use it like this: `xAxis.scale.linear` following suit with how Feliz-style bindings are built

PR #163, opened 2020-04-20
Rename "inheritFromParent" to "inherit"

dbrattli:
I cannot see that "inheritFromParent" is a valid style value. Is this a typo?


Zaid-Ajaj:
Thanks! It was probably a copy-paste error

PR #165, opened 2020-04-21
Update to Fable.React 6.0.0

MangelMaxime:
Fable.React 6.0.0 [removed the type alias for FunctionComponent](https://github.com/fable-compiler/fable-react/pull/188/files#diff-80320e3248fd41bc3ce59f9bc8dc53b4L6).

I am unsure if this is deliberate or something we forget to undo when implementing the new version.

But because it was just an alias, I fixed Feliz code directly.

### DIFF
--- a/Feliz/React.fs
+++ b/Feliz/React.fs
@@ -23,7 +23,7 @@ type internal Internal() =
             ?name: string,
             ?withKey: 'props -> string
         )
-        : Fable.React.FunctionComponent<'props> =
+        : 'props -> Fable.React.ReactElement =
             name |> Option.iter (fun name -> renderElement?displayName <- name)
             fun props ->
                 let props = props |> propsWithKey withKey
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj:
Thanks a lot Maxime :heart: :pray: I have tested it with the docs application and works like a charm. 

I wonder if I have to update the [Common Pitfalls](https://zaid-ajaj.github.io/Feliz/#/Feliz/React/CommonPitfalls) page  where it recommends:
> Components must be defined at the module level 

Now that the main Fable.React has built-in cache 🤔

MangelMaxime:
I think you are not using the API that Alfonso designed.

But, I think it could be a good idea to use Alfonso's API even if you just create a wrapper on top of it because it got nice improvements and also support HMR out of the box.

About the section of the documentation you mentioned, if you use Alfonso's API you could just rephrase your sentence to something like that:

Even though you can define a React component inside another function, we encourage you not to do so. It will help you isolate your code/better structure it.

I shifted the advice to writing "better" code. TBH, I don't think it is a good idea to create a component inside a component unless to have a really good reason for that.

PR #169, opened 2020-04-26
Link does not have an end tag

dbrattli:
The `link`element does not have an end tag and cannot have children.

https://html.spec.whatwg.org/#the-link-element


Zaid-Ajaj:
Nice catch, thanks a lot!

PR #170, opened 2020-04-26
Add missing member on prop for onMouseUp event

Samuel-Dufour:


Zaid-Ajaj:
Hi @Samuel-Dufour, thanks a lot for adding this property! I don't know how we missed it :smile: I have merged and published a new version of Feliz as of v0.77.0 which should be available soon on nuget :rocket:

PR #173, opened 2020-04-28
Add charset property

dbrattli:


Zaid-Ajaj:
Thanks @dbrattli, I think I will also add a static property for the most commonly used charset: UTF-8 like this:
```fs
[<Erase>]
type charset = 
    static member inline utf8 = Interop.mkAttr "charset" "UTF-8"
```
Maybe you can add it to Feliz.ViewEngine as well? :wink:

dbrattli:
Yes, already added to Feliz.ViewEngine where I was actually needing it. Working on a SSR example and running into issues all the time, so if I can get it working then most things should work.

PR #177, opened 2020-05-02
Add React.useCallbackRef hook

Shmew:
This PR proposes an addition of a hook I named `useCallbackRef` which is similar to `React.useCallback`, but allows fetching fresh data without causing additional renders.

You can see an example in the diff of `App.fs`, where it renders a function component with two children, one showing the render count of the parent, and one of the child. Both the parent and child call `renderCount` which uses a ref and adds to the counter which is displayed in a div to show the times the component has been rendered.

The parent component sets an interval to call `setCount` every second, which will cause a re-render of the parent. Since the child is called with `React.memo` it does not re-render from this.

The difference between the two functions that the main function calls, is that one wraps this action in `React.useCallback` and the other in `React.useCallbackRef`. Running the example you can see that while neither page renders additional times, the example using `React.useCallback` always dispatches the alert as `1` while the one using `React.useCallbackRef` always has the current count.

There are two benefits of this:
* No need for dependency array
* No additional renders despite gaining access to that new set of data.

callbackStatic


Zaid-Ajaj:
Merged and published as of v0.82.0 :rocket:

Issue #127, opened 2019-11-26
Finish React API bindings


Shmew:
Some things are missing, such as:
- [x] `React.forwardRef`
- [ ] `React.lazy`
- [x] `React.strictMode`
- [ ] `React.suspense`
- [x] `React.useRef`

Issue #50, opened 2019-09-14
Complete API implementation for hooks from React 


Zaid-Ajaj:
Complete API implementation for hooks from React, see [API Reference](https://reactjs.org/docs/hooks-reference.html)

### Basic Hooks
- [x] `React.useState`
- [x] `React.useEffect` 
- [x] `React.useContext`
### Additional Hooks
 - [x] `React.useReducer`
 - [x] `React.useCallback`
 - [x] `React.useMemo`
 - [x] `React.useRef`
 - [x] `React.useLayoutEffect`
 - [x] `React.useDebugValue`
 - [ ] `React.useImperativeHandle`

Suggestions/ideas/PRs with the implementation of these is very much welcome and highly appreciated 🙏

cmeeren:
See https://github.com/fable-compiler/fable-react/commit/fa43005e82490945e2a59beedf86054ae1e929a3 for a discussion of Forcing `option`-wrapping of the ref hook's `current` value.

PR #179, opened 2020-05-03
Add React.forwardRef

Shmew:
Adds the `React.forwardRef` feature.

I also separated out some of the examples and test functions from the main `App.fs` so it's a bit easier to navigate.

Partial work towards #127
Should also un-block #50 for `useImperativeHandle`


Zaid-Ajaj:
Awesome stuff! Thanks a lot Cody :heart: :pray:

PR #191, opened 2020-05-18
Initial implementation of Feliz.UseDeferred

Zaid-Ajaj:
This is an initial API design for using the `Deferred<'T>` type in a nice way that works with React function components and follows hook rules. It exposes two methods: 
 - `React.useDeferred`
 - `React.useDeferredCallback` 

I have included two live examples in the docs of they should work 

@Shmew What do you think of this API, it is a bit different than the one we discussed before but covers quite a lot of ground for many use cases.


Zaid-Ajaj:
Thanks for the review @Shmew, I have simplified `React.useDeferred` as discussed to return the state variable back to the caller

PR #193, opened 2020-05-21
Provide default fallback for suspense

Shmew:
Makes the suspense usage better when a fallback isn't provided.

suspense


Zaid-Ajaj:
Thanks! I have merged and published the addition as of v1.1.0 🎉

PR #196, opened 2020-05-23
Fix displayName for memoized components

### LINKED ISSUE #181: Component name showing up as Anonymous
I've just updated to Fable.React v6 and am using `Feliz.React.memo (name * areEqual * render)` to create a component. When I check the components in the browser, the name is not propagated. Currently I am unable to use `FunctionComponent.Of`, so I'm not sure if this is a problem in Feliz or in the new version of Fable.React.

Edit: the names are there when I switch to `Feliz.React.functionComponent`

kerams:
Fixes #181.


Zaid-Ajaj:
Thanks a lot @kerams, the fix has been published as of v1.1.1 :rocket:

PR #202, opened 2020-06-03
Add missing properties for outline-style

Shmew:
Source: https://developer.mozilla.org/en-US/docs/Web/CSS/outline-style


Zaid-Ajaj:
Awesome! thanks a lot 🙏

PR #204, opened 2020-06-06
Add seq values

Shmew:
* Fixes Fable.Core dependency restrictions that disallowed 4.7.0 for some libraries
* Adds seq overloads for `value` `defaultValue` and `valueOrDefault` to support things like `<select>` elements with the `multiple`.
* Updated test suite to latest release of RTL


Zaid-Ajaj:
Awesome! Thanks a lot :heart:

PR #212, opened 2020-06-23
Added Solution Items to Template

l3m:
Make package.json and webpack config 1-click
accessible when using Rider.

image


Zaid-Ajaj:
Thanks a lot! this is a nice addition. Should be available as of v2.3.0 :smile:

PR #214, opened 2020-06-24
Add Html.textf

aspnetde:
This one is shamelessly stolen from https://github.com/Zaid-Ajaj/Feliz/pull/213, as I noticed that I am using `sprintf` not only for props, but also for `Html.text`:

```fs
let render state _ =
    Html.div [
        Html.text (sprintf "Hello, %s! Go ahead and " state.Session.User)
        Html.a [ ... ]
        Html.text " your name."
    ]
```

This can now become

```fs
    Html.div [
        Html.textf "Hello, %s! Go ahead and " state.Session.User
        Html.a [ ... ]
        Html.text " your name."
    ]
```

I took over the tests, as it seems to make sense to test for the same cases (although the test arrangement with `Html.textf` within the children of another element is artificial).

### DIFF
--- a/Feliz/Html.fs
+++ b/Feliz/Html.fs
@@ -712,6 +712,8 @@ type Html =
     static member inline text (value: string) : ReactElement = unbox value
     static member inline text (value: System.Guid) : ReactElement = unbox (string value)
 
+    static member inline textf fmt = Printf.kprintf Html.text fmt
+    
     static member inline textarea xs = Interop.createElement "textarea" xs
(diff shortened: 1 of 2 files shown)


Shmew:
Beat me to the punch! I was about to make a PR 😄 

👍

Zaid-Ajaj:
Awesome work @aspnetde, will publish soon once we add another bit of functionality for request cancelation

aspnetde:
That’s fine with me, take your time ✌️

Zaid-Ajaj:
Published and will be available as of v1.5 :rocket: (currently being indexed by nuget)

PR #215, opened 2020-06-24
Add React.useCancellationToken hook

Shmew:
Adds the `React.useCancellationToken` hook to make managing async functions inside components easier to manage.


Zaid-Ajaj:
Published and will be available as of v1.5 🚀 🎉

PR #216, opened 2020-06-26
Fix React state update bug

Shmew:
Prevents a state update when the component does not wait for the update function to finish (such as a `Cmd` itself being the reason for the unmount).

If a component immediately unmounts during a dispatch, React throws `Can't perform a React state update on an unmounted component`. Checking if the token has requested a cancellation prevents this.


aspnetde:
Incredible. I just ran into this with v1.2.1 and was about to open an issue, when I saw that PR. Just in time :-).

PR #217, opened 2020-06-29
Fix template

l3m:
Apparently, .gitignore and .gitattributes are not part of the template, so this is a quick fix to remove the broken references from App.sln.

### DIFF
--- a/Feliz.Template/Feliz.Template.proj
+++ b/Feliz.Template/Feliz.Template.proj
@@ -15,6 +15,8 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
     <PackageReleaseNotes>Update dependencies</PackageReleaseNotes>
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @l3m, any ideas why these files aren't included in the template? I would rather fix the problem by adding them into the template instead because they are kinda important so you don't accidentally publish `.ionide` or something like that

l3m:
I thought they were, that's why I added them initially. 

In the meantime, I found it:
https://github.com/dotnet/templating/issues/2350#issuecomment-610431461

Will update.

l3m:
@Zaid-Ajaj Updated, should work now. Unsure how to test it locally though.

Zaid-Ajaj:
There is one way to find out if it works :wink:

l3m:
Packed the nuget locally and it did contain the .gitignore and .gitattributes files inside Content.

Zaid-Ajaj:
Awesome! It should be published as of v2.5.0 🚀  thanks a lot!

PR #218, opened 2020-07-01
Allow passing parameter to start function of `useDeferredCallback`

johannesegger:
Using a closure might not work because React seems to cache the `operation` function. So in my case the following example prints different values to the console:

```fsharp
let (sendOrderState, setSendOrderState) = React.useState(Deferred.HasNotStartedYet)
let (authKey, setAuthKey) = React.useState(Deferred.HasNotStartedYet)

// ...

let startSendOrder =
    React.useDeferredCallback(
        (fun () -> async {
            printfn "%O" authKey
            // TODO use auth key
        }),
        setSendOrderState
    )

// Chain two `Deferred`s together - if `authKey` is resolved and `sendOrderState` hasn't started, start `sendOrderState`
if Deferred.resolved authKey && Deferred.hasNotStartedYet sendOrderState then
    printfn "%O" authKey
    startSendOrder ()
```

This PR enables passing `authKey` as parameter to `startSendOrder` which obviously results in the same values printed to the console.

It could easily be possible that I'm doing it all wrong - I just started with Feliz and React hooks.
I would appreciate any advice in this case.

### DIFF
--- a/Feliz.UseDeferred/UseDeferred.fs
+++ b/Feliz.UseDeferred/UseDeferred.fs
@@ -79,27 +79,27 @@ module ReactHookExtensions =
 
             deferred
 
-        static member useDeferredCallback(operation: unit -> Async<'T>, setDeferred: Deferred<'T> -> unit) =
+        static member useDeferredCallback(operation: 'TIn -> Async<'TOut>, setDeferred: Deferred<'TOut> -> unit) =
             let cancellationToken = React.useRef(new System.Threading.CancellationTokenSource())
-            let executeOperation = async {
+            let executeOperation arg = async {
                 try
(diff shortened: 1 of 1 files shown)


Shmew:
That solution will work with a normal input, but (I believe) the issue will persist if the input is a function. I think a solution for this is to make the hook inline so we can call `useCallbackRef` on the input operation function.

Zaid-Ajaj:
@johannesegger I think it looks better that way since it is more generic and doesn't require closure mechanism to cache the input argument. I've merged this PR and updated `Feliz.UseDeferred` to v1.1.1 :rocket: so it should be available soon after nuget does its magic

PR #221, opened 2020-07-03
Adds support for IDisposable option overloads

Shmew:
There's a somewhat common pattern that when working with objects than interface `System.IDisposable` asynchronously that to properly make sure they get disposed when the element unmounts that they're stored as a `IRefValue<System.IDisposable option>`. 

This is a bit annoying to work with as you need to do: 
```fs
React.createDisposable(fun () -> myRef.current |> Option.iter (fun myDisposable -> myDisposable.Dispose()))
``` 

This PR adds an overload such that you can just return `myRef.current` at the end of your effect.

As you can see from the diff of `useElmish`:
image

### DIFF
--- a/Feliz.UseElmish/UseElmish.fs
+++ b/Feliz.UseElmish/UseElmish.fs
@@ -83,10 +83,7 @@ module UseElmishExtensions =
 
             let dispatch = React.useCallbackRef(dispatch)
 
-            React.useEffectOnce(fun () ->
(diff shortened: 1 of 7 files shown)


Zaid-Ajaj:
Awesome stuff, this keeps getting better :heart:

Zaid-Ajaj:
Published as of v1.6 :rocket:

PR #223, opened 2020-07-04
Bugfix: Safari UseMediaQuery

l3m:
Moved the UseMediaQuery hook to the old version as
a workaround for missing functionality on Safari.

### DIFF
--- a/Feliz.UseMediaQuery/UseMediaQuery.fs
+++ b/Feliz.UseMediaQuery/UseMediaQuery.fs
@@ -5,6 +5,9 @@ open Browser
 open Browser.Types
 open Feliz
 
+open Fable.Core
+open Fable.Core.JsInterop
+
 [<RequireQualifiedAccess>]
 type ScreenSize =
     | Mobile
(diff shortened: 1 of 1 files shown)


Feliz.UseMediaQuery/UseMediaQuery.fs
```diff
@@ -38,26 +41,38 @@ let private makeQueries breakpoints =
...
+                handler ()
+                addListener mediaQueryList handler
 
-            , [| query :> obj |])
+                let ret =
+                    {new IDisposable with
+                        member this.Dispose() =
+                            removeListener mediaQueryList handler}
+                ret
+            , [| setMq :> obj; mediaQuery :> obj |])
```
Zaid-Ajaj, line 70: I am not that `setMq` itself can be a dependency here and it doesn't really change

l3m: Indeed. Fixed.

Zaid-Ajaj:
Published as v1.0.2 with the fix included and FSharp.Core downgraded to v4.7.0 (which we should probably do for Feliz as well)

PR #224, opened 2020-07-04
Move to paket, add linting

Shmew:


Feliz.sln
```diff
@@ -19,24 +19,11 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Feliz.Popover", "Feliz.Popo
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Tests", "tests\Tests.fsproj", "{74AC6D90-9BF7-44B7-ACDD-6C7616CE07E7}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{2979FCC6-85A7-4C95-B05B-D9A5F48B3C6A}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "web", "web", "{2979FCC6-85A7-4C95-B05B-D9A5F48B3C6A}"
```
l3m, line 22: I prefer the original version with 1 folder with these files...

Zaid-Ajaj:
Any ideas why the build is failing? I am guessing it could be due to FAKE version mismatch

Shmew:
There was a wonky merge issue that was fixed. The only problem now is the puppeteer tests are failing due to timeout... it works fine with `npm run test:live`

Zaid-Ajaj:
I want to revisit this soon. I am not a big fan of the fact that we are using nuget and paket at the same time and their config is kinda fighting with each other. I want to simplify all these things and use only one package management system: **nuget**. I know many people are fans of paket but to me it complicates things more than it solves. I also want to move the `build.fsx` file into a console application (that uses nuget) following what I am doing with [Snowflaqe](https://github.com/Zaid-Ajaj/Snowflaqe) which makes things really simple

Shmew:
Yeah sure thing, I didn't delete the branch so you can just clone it if/when you want to mess with it. It was just getting so out of date I figured I should go ahead and close it for now.

PR #234, opened 2020-07-25
Added Instructions to Run Application

iwasdavid:
When I cloned the repo, and was looking for instructions to run the app I couldn't find any in the documentation. Added this section for people not familiar with `npm`.

If these instructions are somewhere that I missed, just close this PR :)

### DIFF
--- a/public/Feliz/ProjectTemplate.md
+++ b/public/Feliz/ProjectTemplate.md
@@ -22,6 +22,19 @@ dotnet new -u Feliz.Template
 dotnet new -i Feliz.Template
 ```
 
+### Run the app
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Thanks a lot for this addition, the docs should now be updated. The instructions to run the application are actually on the `README.md` file that is created when you create the template using `dotnet new feliz` but it doesn't hurt to have the intructions on the docs page too :smile:

iwasdavid:
Ah I see. I should have looked there haha. Hopefully it helps one other person then it was worth it 👍

PR #243, opened 2020-08-07
React.delay and React.delaySuspense components

Shmew:
This PR adds an abstraction over a relatively common use case of delaying rendering for set periods of time (typically loading components).

See the docs for the finished api.

Here's the idea in action:

delayed

### DIFF
--- a/Feliz.Delay/Feliz.Delay.fsproj
+++ b/Feliz.Delay/Feliz.Delay.fsproj
@@ -1,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageVersion>0.1.0</PackageVersion>
+    <Authors>Cody Johnson, Zaid Ajaj</Authors>
+    <Description>Adds easy to use delayed rendering</Description>
+    <PackageLicenseUrl>https://github.com/Zaid-Ajaj/Feliz/blob/master/LICENSE</PackageLicenseUrl>
+    <RepositoryUrl>https://github.com/Zaid-Ajaj/Feliz</RepositoryUrl>
+    <PackageTags>fsharp;fable;react;html;feliz</PackageTags>
+    <PackageReleaseNotes>Initial release</PackageReleaseNotes>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="*.fsproj; *.fs; *.js;" PackagePath="fable\" />
+    <Compile Include="Delay.fs" />
+    <Compile Include="Templates.fs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Fable.Browser.MediaQueryList" Version="1.1.0" />
+  </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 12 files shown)


Zaid-Ajaj:
Hi Cody, I really like the APIs here, especially `React.delay` (nice and simple). However, I don't think I want to add all of these helper functions into the base Feliz library. I know it makes things simpler to put it all in here but it clutters the API a lot if you are not going to use any of these.  A standalone small library seems like the better option to me. 

> `React.delay` sounds like a side-effectful function 🤔 maybe rename to `React.delayElement` (it's up you, of course)

Shmew:
Yeah, that's what I realized as I fleshed things out. Do you want me to create a project here (the number of libraries in this repo is already getting pretty big though), or I can create a separate repo. Another option and something I think we've discussed before was merging some of the smaller libraries that have no dependencies into the base library, but add additional namespaces so they're still opt-in without having to grab many different packages. Namely `Feliz.UseDeferred` and `Feliz.UseMediaQuery`.

Zaid-Ajaj:
> Do you want me to create a project here (the number of libraries in this repo is already getting pretty big though)

For now, yes please! :pray: currently I don't mind the number of packages in the repo. However, I would like to explore the other option in the next big release (v2.0) "was merging some of the smaller libraries that have no dependencies into the base library". Following this logic, we might even add Feliz.Router and Feliz.UseElmish along with other smaller packages as separate namespacing so that users literally get the "full package" when they install Feliz. 

I think it would be interesting, no one likes to install and maintain many packages, especially those you expect people to use (especially Feliz.Router). 

First I am waiting for the Fable plugins to come out so we can start improving the code-gen and supporting fast-refresh, I am really excited about those

Shmew:
> For now, yes please! 🙏 currently I don't mind the number of packages in the repo.

Cool, I'll get that setup, tests, and add docs later today.

 > Following this logic, we might even add Feliz.Router and Feliz.UseElmish along with other smaller packages as separate namespacing so that users literally get the "full package" when they install Feliz.

I think `Feliz.UseListener` would be another one that could merge into the base library. Certainly would be nice to clean up all the single file libraries we have floating around.

Feliz.Delay/Feliz.Delay.fsproj
```diff
@@ -0,0 +1,30 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="*.fsproj; *.fs; *.js;" PackagePath="fable\" />
+    <Compile Include="Delay.fs" />
+    <Compile Include="Templates.fs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Fable.Browser.MediaQueryList" Version="1.1.0" />
```
Zaid-Ajaj, line 21: The dependency `Fable.Browser.MediaQueryList` can be removed here, right?

Feliz.Delay/Feliz.Delay.fsproj
```diff
@@ -0,0 +1,30 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="FSharp.Core" Version="4.7.2" />
```
Zaid-Ajaj, line 24: I think it is OK if we use `4.7.0` as a minimum version for FSharp.Core like the other components are doing

PR #245, opened 2020-08-07
Add units to translate functions

zanaptak:
The `translate*` functions fail to render with non-zero integer values, i.e. `style.transform.translateX 0` will render the style, but `style.transform.translateX 1` won't.

The values are defined as `<length-percentage>` values on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/translate).

This PR adds `ICssUnit` overloads, and adds `px` to the original `int` overloads.


Zaid-Ajaj:
Thanks a lot @zanaptak for the fix, it should now be included in the Feliz v1.11.0 :rocket: and will available in a bit

PR #246, opened 2020-08-08
Add float to styles

Shmew:
Adds the `float` CSS property to `style`

Sourced from [here](https://developer.mozilla.org/en-US/docs/Web/CSS/float).


Zaid-Ajaj:
Thanks a lot Cody! this addition along with font display properties have been added to the v1.11.0 release and should now be available.

> BTW, I removed Feliz.ElmishComponents from the docs in favor of Feliz.UseElmish and will probably remove the project entirely in the next major release

PR #247, opened 2020-08-09
Remove usage of sprintf

Shmew:
There were still a few lingering instances where we use `sprintf` that needed to be cleaned up.


Zaid-Ajaj:
Looks great!

Zaid-Ajaj:
Published the goodies :rocket:

Issue #238, opened 2020-07-29
Validity of Html.text list overload?


zanaptak:
`Html.text` has an overload taking a list parameter and creating a `<text>` element.

https://github.com/Zaid-Ajaj/Feliz/blob/a0e39f6cf2fa411f25bea524b6db05d8d3a2edfe/Feliz/Html.fs#L709

Should this overload exist?

As far as I can tell, `<text>` is not a valid HTML element ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element), [Spec](https://www.w3.org/TR/html52/), [Validator](https://validator.w3.org/)).

I imagine it works by accident in practice due to browser leniency treating it as a div.

I'm wondering if Html.text should only allow scalar values and serve strictly as a shortcut for bare text, rather than a non-standard element. And users wanting to add props or children would have to explicitly change it to div/span or something.

Zaid-Ajaj:
I believe this overload is for [SVG text](https://www.w3schools.com/graphics/svg_text.asp) not just HTML

PR #249, opened 2020-08-11
Fix prop.transform for Svg elements

lordnull:
Feliz.transform.translate is an example where this change is useful. The
function accepts `ICssUnit * ICssUnit`. However, the SVG spec does not
expect a unit for the numbers given in the translate part of a
transformation. This allows the api to remain the same (no sudden change
on what is expected), but outputs what is need to have transformations work.

### DIFF
--- a/Feliz/Length.fs
+++ b/Feliz/Length.fs
@@ -84,7 +84,3 @@ type length =
     static member inline auto : ICssUnit = unbox "auto"
     /// calculated length, frequency, angle, time, percentage, number or integer
     static member inline calc(value:string) : ICssUnit = unbox ("calc(" + (unbox<string>value) + ")")
-    /// There are times a unit is not appropriate, like an svg's translate.
-    static member inline nounit(value:int) : ICssUnit = unbox (unbox<string>value)
-    /// There are times a unit is not appropriate, like an svg's translate.
-    static member inline nounit(value:double) : ICssUnit = unbox (unbox<string>value)
--- a/Feliz/Length.fs
+++ b/Feliz/Length.fs
@@ -84,3 +84,7 @@ type length =
     static member inline auto : ICssUnit = unbox "auto"
     /// calculated length, frequency, angle, time, percentage, number or integer
     static member inline calc(value:string) : ICssUnit = unbox ("calc(" + (unbox<string>value) + ")")
+    /// There are times a unit is not appropriate, like an svg's translate.
+    static member inline nounit(value:int) : ICssUnit = unbox (unbox<string>value)
+    /// There are times a unit is not appropriate, like an svg's translate.
+    static member inline nounit(value:double) : ICssUnit = unbox (unbox<string>value)
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -1770,7 +1770,10 @@ type prop =
         Interop.mkAttr "transform" (unbox<string> transform)
     /// Defines a list of transform definitions that are applied to an element and the element's children.
     static member inline transform (transforms: seq<ITransformProperty>) =
-        Interop.mkAttr "transform" (unbox<seq<string>> transforms |> String.concat " ")
+        let unitList = [ "px" ; "deg" ]
+        let removeUnits (s : string) =
+            List.fold (fun (ins:string) toReplace -> ins.Replace(toReplace,"")) s unitList
+        Interop.mkAttr "transform" (unbox<seq<string>> transforms |> Seq.map removeUnits |> String.concat " ")
 
     /// Sets the `type` attribute for the element.
     static member inline type' (value: string) = Interop.mkAttr "type" value
--- a/Feliz/Transform.fs
+++ b/Feliz/Transform.fs
@@ -24,7 +24,12 @@ type transform =
     /// Defines a 2D translation.
     static member inline translate(x: int, y: int) =
         unbox<ITransformProperty> (
-            "translate(" + (unbox<string> x) + "px," + (unbox<string> y) + "px)"
+            "translate(" + (unbox<string> x) + ", " + (unbox<string> y) + ")"
+        )
+    /// Defines a 2D translation.
+    static member inline translate(x: float, y: float) =
+        unbox<ITransformProperty> (
+            "translate(" + (unbox<string> x) + ", " + (unbox<string> y) + ")"
         )
     /// Defines a 2D translation.
     static member inline translate(x: ICssUnit, y: ICssUnit) =
@@ -35,7 +40,12 @@ type transform =
     /// Defines a 3D translation.
     static member inline translate3D(x: int, y: int, z: int) =
         unbox<ITransformProperty> (
-            "translate3d(" + (unbox<string> x) + "px," + (unbox<string> y) + "px," + (unbox<string> z) + "px)"
+            "translate3d(" + (unbox<string> x) + ", " + (unbox<string> y) + ", " + (unbox<string> z) + ")"
+        )
+    /// Defines a 3D translation.
+    static member inline translate3D(x: float, y: float, z: float) =
+        unbox<ITransformProperty> (
+            "translate3d(" + (unbox<string> x) + ", " + (unbox<string> y) + ", " + (unbox<string> z) + ")"
         )
     /// Defines a 3D translation.
     static member inline translate3D(x: ICssUnit, y: ICssUnit, z: ICssUnit) =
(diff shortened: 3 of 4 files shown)


Zaid-Ajaj:
Hi @lordnull, 

I have thought about the issue this morning and although this change fixes the problem, I would rather fix the function overload itself and allow it to accept `int * int` (or other variants) instead of a no-unit unit. This is also because we are planning on moving away from `ICssUnit` and instead use units of measure but we are not there yet.

Can you please add the expected method overloads to the `transform` function instead?

lordnull:
Thanks for taking a look @Zaid-Ajaj !

I've implemented the requested changes. However, this does change the api a bit. When doing a style transfer that takes `ITransformProperty list`, if `int` or `float` is used for the translations, the result will no longer have a unit. This breaks HTML. The inverse, keeping a unit by default, breaks SVG.

I used a work-around in my code before doing the pr of defining an a couple of functions that return an `CssUnit` with no unit on the output string, so this is starting to feel like replacing one paper cut with another.

In the long term, it may be worth-while to split SVG into a space separate from HTML, as there are plenty more parts where the specs diverge, resulting in surprising behavior for HTML users or SVG users.

zanaptak:
I've only recently added the `px` suffixes to the int overloads (#245), similar to e.g. the margin int overloads, to make them valid HTML. Unfortunately I didn't consider SVG. Given that, I feel you are correct to remove the implicit `px` as at least the length overloads are now available.

I've been caught out by HTML/SVG overlap before (#238), I too wonder if a rethink is warranted.

Also wondering if all implicit `px` should be removed. Make units or lack thereof explicit, expectation on user to know HTML/SVG context and the appropriate units allowed.

Zaid-Ajaj:
I remember a while ago @Shmew had the same discussion with SVG attributes vs CSS styles and we decided to give CSS priority when there is a property in CSS that works the same as the corresponding SVG property. 

Following that logic, we can say that this isn't really an issue and @lordnull is better off using the `transform` style rather than the `transform` attribute which should output the implicit `px` unit as defined in the CSS spec. But then `prop.transform` wouldn't work because the HTML attribute explicitly wants no units at all just like @zanaptak was saying. 

Hmm, I will have to think about this one, at first I didn't really want separate namespaces with `Html` and `Svg`, that's why it is all "unified" but apparently this can cause such problems. 

If you have suggestions @zanaptak @Shmew I would love to hear them. I know Cody was looking into implementing units of measure overloads for certain functions such as:
```fs
style.margin 5<px>
``` 
This might actually solve the ambiguity problem but again we *DO* want some overloads to equal the same thing like `style.margin 5<px>` and `style.margin 5` 🤔

Zaid-Ajaj:
Another option of course is to remove `prop.transform` altogether and instead only rely on the `style` to make use of transforms. This way we get rid of ambiguity and have one way to do things

Shmew:
I'm a fan of having `Svg` and `Html` so there's no ambiguity. 

>  I know Cody was looking into implementing units of measure overloads

So far just an idea I want to try, but I plan to jump into it when I get a bit more time. As far as I can tell it should work fine.

zanaptak:
Actually since there is distinct behavior here, really it seems like there should be separate ancillary types, say `css.transform.*` and `svg.transform.*` with independent overloads as appropriate.

In general I think separate namespaces overall would be a better user experience, and should ideally include props, but I don't know what that would look like. `Svg` and `svgprop`? `Svg.elem.*` and `Svg.prop.*`? There would be better discoverability and safety in each context and less clutter of invalid properties.

Disclaimer: I have very little SVG experience, which may color my opinions.

zanaptak:
Anybody know of any good examples of SVG-in-Feliz to get an idea how it looks in practice currently?

Zaid-Ajaj:
@zanaptak The [Feliz.Recharts docs](https://zaid-ajaj.github.io/Feliz/#/Recharts/Overview) have a bunch of those but they are quite simple examples:
```fs
let getGradientDefinition (gradientId: string) (data: Point list) =
    let offset = getOffset data
    Html.defs [
        Html.linearGradient [
            prop.id gradientId
            prop.x1 0; prop.x2 0
            prop.y1 0; prop.y2 1
            prop.children [
                Html.stop [
                    prop.offset offset
                    prop.stopColor "green"
                    prop.stopOpacity 1.0
                ]

                Html.stop [
                    prop.offset offset
                    prop.stopColor "red"
                    prop.stopOpacity 1.0
                ]
            ]
        ]
    ]
```
And
```fs
let renderCustomLabel (input: IPieLabelProperties) =
    let radius = input.innerRadius + (input.outerRadius - input.innerRadius) * 0.5;
    let radian = System.Math.PI / 180.
    let x = (input.cx + radius * cos (-input.midAngle * radian))
    let y = (input.cy + radius * sin (-input.midAngle * radian))

    Html.text [
        prop.style [
            style.fill color.white
        ]
        prop.x x
        prop.y y
        prop.dominantBaseline.central
        if x > input.cx then prop.textAnchor.startOfText else prop.textAnchor.endOfText
        prop.text (sprintf "%.0f%%" (100. * input.percent))
    ]
```
The second example shows the "problem" where instead of `prop.fill` we are using the `prop.style [ style.fill whatever ]` and now that I am looking at it, it's a bit (unnecessarily) ugly. @Shmew what do you think?

lordnull:
I don't have a lot of svg in Feliz experience, but I do have plenty in elm. Elm has both an `Html` and `Svg` module, duplicating the property names as needed, but keeping the pattern of use the same (since they are both xml related).

So, @Zaid-Ajaj , in the example you posted, swap the Html for Svg and you'll have (at least what is to me) a familiar and expected api for modules that only need Svg.

For modules that need both Html and Svg, using qualified names has worked for me.

Just my 2¢.

Shmew:
I'm no SVG expert, but my impression (which may be wrong) was that it's encouraged to use CSS when possible. It would be nice if we could have `prop.fill` and it apply it to the style. I don't think something like that is really possible without performance hits, or if the plugin system works for something like that in Fable 3.

A couple options are doing things like element typed properties like `stop.offset`, or even just `svg.offset` rather than `prop.offset`.

Zaid-Ajaj:
Thanks @lordnull and @Shmew for the input

implementing a `Svg` module and its corresponding properties into another `svg` module sounds nice to me. This is the same structure we follow with third-party APIs and it will look like this:
```fs
let renderCustomLabel (input: IPieLabelProperties) =
    let radius = input.innerRadius + (input.outerRadius - input.innerRadius) * 0.5;
    let radian = System.Math.PI / 180.
    let x = (input.cx + radius * cos (-input.midAngle * radian))
    let y = (input.cy + radius * sin (-input.midAngle * radian))

    Svg.text [
        svg.x x
        svg.y y
        svg.fill.white
        svg.dominantBaseline.central
        if x > input.cx then svg.textAnchor.startOfText else svg.textAnchor.endOfText
        svg.text (sprintf "%.0f%%" (100. * input.percent))
    ]
```
And
```fs
let getGradientDefinition (gradientId: string) (data: Point list) =
    let offset = getOffset data
    Svg.defs [
        Svg.linearGradient [
            svg.id gradientId
            svg.x1 0; svg.x2 0
            svg.y1 0; svg.y2 1
            svg.children [
                Svg.stop [
                    svg.offset offset
                    svg.stopColor "green"
                    svg.stopOpacity 1.0
                ]

                Svg.stop [
                    svg.offset offset
                    svg.stopColor "red"
                    svg.stopOpacity 1.0
                ]
            ]
        ]
    ]
```
Then going back to the original issue of `transform`, it might look something like this:
```fs
svg.transform [
    svg.transform.translateX(10)
    svg.transform.translateY(20)
]
```
@lordnull @Shmew @zanaptak How does this look like to you?

zanaptak:
Pending final decision on a wider SVG effort, any thoughts on this PR in the meantime? Without it, `prop.transform` has unexpectedly broken for SVG users as of 1.11. And reverting to pre-1.11 state for `style.transform` seems unlikely to affect many users, possibly just me, since it has only been updated for a week.

Zaid-Ajaj:
@zanaptak I created issue #251 to properly implement the Svg modules. As for this PR, I don't it fixes the issue at hand (sorry @lordnull ) because current changes break `style.transform` and fix `prop.transform` while actually we want both to work. 

For the time being, I still want to fix this issues *without* breaking changes and make both `style.transform` and `prop.transform` work. I am thinking of a much simpler *hack* until the SVG issue is resolved and it goes as follows: inside the implementation of `prop.transform`, replace "px" with an empty string if it occurs (potentially other units as well except for percentages?) :joy: yeah I know it is hacky but it ensures there are no breaking changes, both variants will work *and* the perf impact is insignificant

lordnull:
I'll give a go at updating the tests/code @Zaid-Ajaj .

Zaid-Ajaj:
Published and included as of v1.13 :rocket:

PR #250, opened 2020-08-15
Add ReactDOM.createPortal

fracek:
I added a binding to `ReactDOM.createPortal` since it was missing, together with an example on how to use it.

Let me know if you think I should change anything.


Zaid-Ajaj:
This is awesome @fracek thanks a lot for the PR, the tests and the documentation :heart: 😍

Zaid-Ajaj:
Published and included as of v1.13 🚀

Issue #263, opened 2020-10-08
Question: Feliz.ElmishComponents vs Feliz.UseElmish


Dzoukr:
Hello friend, it's been a while I bothered you with some retarded question, so it's time for another take. 😂 

Can you please somehow describe what is the difference between `Feliz.ElmishComponents` and `Feliz.UseElmish`? I see them both are React component having Elmish MVU parts. Are there any typical scenarios where you would choose one instead of another? I love easy of use of `Feliz.UseElmish` and just fear I am maybe missing something even cooler here. 😄

Shmew:
So `Feliz.ElmishComponents` isn't very useful anymore, and predates `Feliz.UseElmish`. The reason for this is that originally `Feliz.ElmishComponents` was a class based component, which I rewrote as a function component. A bit after that, we realized that if it's a function component it's trivial to just convert it into a hook. That is how `Feliz.UseElmish` came to be. So now `Feliz.UseElmish` is really just a function component that calls the hook.

Dzoukr:
Thanks a lot, Cody!!! So if I get i correctly, `Feliz.UseElmish` is the default one, right now.

Shmew:
Yeah, we mostly still have `Feliz.ElmishComponents` around to not break people. You can consider it to be deprecated.

Dzoukr:
Perfect! Thanks again!

MangelMaxime:
If it is "deprecated" it could be a good idea to mark it "Obselete" and redirect people to the recommended API.

What do you think?

Dzoukr:
Yes, we are just discussing with colleagues from F# team that we should do PR to documentation (maybe with link to this issue)

Dzoukr:
Ok, I did it here - https://github.com/Zaid-Ajaj/Feliz/pull/264

Zaid-Ajaj:
> Are there any typical scenarios where you would choose one instead of another? I love easy of use of Feliz.UseElmish and just fear I am maybe missing something even cooler here. 😄

Hi Roman, good question! 

The difference like Cody mentioned is that `Feliz.UseElmish` is written with hooks as opposed to a full component implementation of `Feliz.ElmishComponents`. 

[...]

PR #264, opened 2020-10-08
ElmishComponents deprecation note

Dzoukr:
Based on this https://github.com/Zaid-Ajaj/Feliz/issues/263#issuecomment-705370349 I would put a note in official docs


Zaid-Ajaj:
Thanks :smile:

Dzoukr:
I cannot see it in official docs - should I do anything else? Was not sure where to change it to be honest. :)

Zaid-Ajaj:
The PR is good, I just didn't publish it from my local machine yet :smile:

PR #265, opened 2020-10-18
Fix some invalid cases in html properties

Titaye:
Fix Invalid DOM property `minlength`. Did you mean `minLength`?
Fix Invalid DOM property `playsinline`. Did you mean `playsInline`?


Zaid-Ajaj:
Thanks a lot @Titaye, the fix should be included in Feliz v.1.14.1 :rocket:

Issue #157, opened 2020-04-10
Fast Refresh


Shmew:
The React team is currently in the process of implementing a native HMR solution which deprecates things like [react-hot-loader](https://github.com/gaearon/react-hot-loader) called [Fast Refresh](https://github.com/facebook/react/issues/16604#issuecomment-528663101). 

I figured we should add an issue to keep this on the radar.

There are a some benefits to be gained from this:
 * Feliz apps won't need to use two different styles in the entry point of the application. 
 * Preserves component state.
 * Doesn't destroy the DOM.
 * Not dependent on webpack
 * Works with React-Native.
 * Faster refreshing (if they live up to their name 😄)

In current state there are some logistics that need to be figured out:

 * Code block that must happen before anything else that injects some `window` variables.
 * All modules that export react components need to be wrapped in a `try .. finally` block and some other things to do with Fast Refresh.

MangelMaxime:
IHMO if possible we should add support for it in Elmish.Debugger because that's the current library responsible for handling HMR.

Feliz doesn't know about the application state when used with Elmish.

It's also important to note that Elmish.Debugger will allow us to support any Elmish application like pure Fable.React, pure Feliz, a mix of both or even just pure Elmish with any renderer.

Shmew:
Yeah that sounds fine to me, I wasn't sure if `Elmish.Debugger` or `Elmish.HMR` were necessarily the right places, I hadn't looked to see if they were directly coupled with React or not.

MangelMaxime:
Oops it's `Elmish.HMR` and not `Elmish.Debugger` I edited my previous message.

No, it's not coupled to React, `Elmish.HMR` shadows some of the Elmish features in order to make them support HMR.

For example:

- `Program.run` from Fable.Elmish
- `Program.toNavigable` from Fable.Elmish.Browser 
- `Program.withReactBatched ` from Fable.Elmish.React
- etc.

Zaid-Ajaj:
@MangelMaxime AFAIL `Elmish.HMR` doesn't handle function components from React and the state is reset when you use them. I will try to make a repro soon

MangelMaxime:
@Zaid-Ajaj From HMR point of view, that's totally normal.

I already have an issue tracking the situation. 

What you are describing is something well know in JavaScript and it's also one of the main problems of HMR. If you want to have HMR in your project you need to make your code understand it. This is not something that comes from nowhere.

[...]

PR #266, opened 2020-10-20
Intial implementation of [<ReactComponent>] and other compiler plugins

Zaid-Ajaj:
> Fable 3 goodness incoming :heart:

Using `[<ReactComponent>]` we will able to properly compile function components in a way that idiomatic to React. This allows tooling around React to work like it should with Feliz. 

The problem with `React.functionComponent` is that it combines the definition of the component _and_ its creation in one go. This doesn't work well with the fast-refresh plugin #157 and many other tools. Instead, the `[<ReactComponent>]` attribute will allow us to write functions which define the components, then when the functions are called, they are transformed at call-site into `createElement` calls from React.
```fs
[<ReactComponent>]
let counter(initial: int) =
    let (count, setCount) = React.useState initial
    Html.div [
        Html.button [
            prop.style [ style.marginRight 5 ]
            prop.onClick (fun _ -> setCount(count + 1))
            prop.text "Increment"
        ]

        Html.button [
            prop.style [ style.marginLeft 5 ]
            prop.onClick (fun _ -> setCount(count - 1))
            prop.text "Decrement"
        ]

        Html.h1 count
    ]
```
compiles to
```js
export function Counter(counterProps) {
  const initial = counterProps.initial
  ...
}
```
This is an idiomatic React component definition. Then on call-site when want to initialize the component like this in F#
```fs
counter 10
```
The call is transformed into the following JS
```js
createElement(Counter, { initial: 10 })
```
Notice the automatic prop object creation and the capitalization of the component as it is the convention in React apps. 

Thanks a lot to @alfonsogarciacaro who is helping me out with juggling the AST and adding more features to Fable to make it work nicely.


Zaid-Ajaj:
Besides `[<ReactComponent>]`, there is work in progress to implement other helper plugins which will be used by Feliz internally. For example `[<IntrinsicElementPrimitive(tagName)>]` will be used on overloads for intrinsic elements (like `div`, `h1`, `span`, etc.) to quickly generate the proper JS without runtime interop:
```fs
// definition
[<IntrinsicElementPrimitive "h1">]`
static member h1 (value: int) = jsNative

// call-site
h1 120

// compiler output
createElement("h1", null, 120)
```
There is also another one in the making called `[<IntrinsicElementChildren>]` which attempts to spread child elements, again without runtime interop
```fs
Html.div [ Html.h1 120 ]
// should generate
createElement("div", null, createElement("h1", null, 120))
```
but that is still WIP

Feliz.CompilerPlugins/ReactComponent.fs
```diff
@@ -0,0 +1,71 @@
...
+            else if decl.Args.Length = 1 && decl.Args.[0].Type = Fable.Type.Unit then
+                // remove arguments from functions requiring unit as input
+                { decl with Args = [ ]; Name = AstUtils.capitalize decl.Name }
+            else
+            // rewrite all other arguments into getters of a single props object
+            // TODO: transform any callback into into useCallback(callback) to stabilize reference
+            let propsArg = AstUtils.makeIdent (sprintf "%sProps" decl.Name)
+            let body =
+                ([], decl.Args) ||> List.fold (fun bindings arg ->
+                    // TODO: detect usage of "key" and emit warning
```
alfonsogarciacaro, line 61: I use this function to detect if an ident is being used in an expression (note you've to use `arg.Name` in case the name has been mangled): https://github.com/fable-compiler/Fable/blob/5690e334d53c0b985a5bbb4d6e141b305ceeec2b/src/Fable.Transforms/FableTransforms.fs#L165-L168

If we don't find `key` then we can just skip the `let key = $props.key` binding.

Feliz.CompilerPlugins/IntrinsicElementPrimitive.fs
```diff
@@ -0,0 +1,18 @@
...
+
+type IntrinsicElementPrimitiveAttribute(tagName:string) =
+    inherit MemberDeclarationPluginAttribute()
+    override _.FableMinimumVersion = "3.0"
+
+    override _.TransformCall(logger, memb, expr) =
+        match expr with
+        | Fable.Call(callee, info, typeInfo, range) when info.Args.Length = 1 ->
+            let element = AstUtils.makeStrConst tagName
+            AstUtils.makeCall (AstUtils.makeImport "createElement" "react") [element; AstUtils.nullValue; info.Args.[0]]
```
alfonsogarciacaro, line 14: You should be able to get the same effect just by inlining, this is what Fable.React does.

Feliz.CompilerPlugins/ReactComponent.fs
```diff
@@ -0,0 +1,71 @@
...
+
+    /// <summary>Transforms call-site into createElement calls</summary>
+    override _.TransformCall(logger, memb, expr) =
+        let membArgs = memb.CurriedParameterGroups |> List.concat
+        match expr with
+        | Fable.Call(callee, info, typeInfo, range) when List.length membArgs = List.length info.Args ->
+            let callee =
+                match callee with
+                | Fable.Expr.IdentExpr ident -> Fable.Expr.IdentExpr { ident with Name = AstUtils.capitalize ident.Name }
+                | _ -> callee
```
alfonsogarciacaro, line 19: If we transform the declaration name, here we should also check if this is an import (when the component is in another file) and capitalize the `Selector` property. Although I would personally prefer if user just capitalize components by themselves as a convention (to distinguish them from just "helper" functions) 😄

Feliz.CompilerPlugins/ReactComponent.fs
```diff
@@ -0,0 +1,71 @@
...
+                // JSX <Component />
+                // JS createElement(Component, null)
+                AstUtils.makeCall (AstUtils.makeImport "createElement" "react") [callee; AstUtils.nullValue]
+            else
+            let propsObj =
+                List.zip membArgs info.Args
+                |> List.choose (fun (arg, expr) -> arg.Name |> Option.map (fun k -> k, expr))
+                |> AstUtils.objExpr
+
+            // TODO: Detect usage of an input called `children` anc spread it into the component
```
alfonsogarciacaro, line 37: At the moment this is complicated because the argument info is not included in the `Fable.Call` expression. In the next update probably we should include the MemberInfo (or parts of it) in CallInfo.

Feliz.CompilerPlugins/IntrinsicElementPrimitive.fs
alfonsogarciacaro: Scratch that, we're going to need the plugin if we want to spread `children`.

alfonsogarciacaro (commented):
Awesome!

Feliz.CompilerPlugins/ReactComponent.fs
Zaid-Ajaj: Sounds like a plan :)

Feliz.CompilerPlugins/IntrinsicElementPrimitive.fs
Zaid-Ajaj: Yeah I am using this attribute specifically for when we have primitive values as the only children of the component:
```
Html.h1 "title" => <h1>title</h1> => createElement("h1", null, "title")
```
Spreading child components across `createElement` is a lot more complicated and it is what `[<IntrinsicElementChildren>]` is trying to accomplish at the moment. However, I am hitting some roadblocks when it comes to lists of components that have non-constant expressions like a conditional in the middle which changes the entire list expression into an ugly delayed sequence, wrapping the entire thing within a `List_fromSeq` function 😢

It is understandable from Fable perspective but it makes spreading the child components not always possible at compile-time so when the basic list pattern isn't detected to the apply the spread, the children are being constructed at runtime as a property

That is also why I can't inline, I want to apply the function body as is without transformation when I can't detect a certain pattern, this way the library stays working even when used from Fable 2.x

Feliz.CompilerPlugins/ReactComponent.fs
Zaid-Ajaj: That would be great to have!

Feliz.CompilerPlugins/ReactComponent.fs
Zaid-Ajaj: > Although I would personally prefer if user just capitalize components by themselves as a convention (to distinguish them from just "helper" functions) 😄

Chances are that this is rather easy to forget and goes against F# style guidelines to define functions within a module in camelCase format, so if it is relatively easy to do then why not :smile:

Feliz.CompilerPlugins/IntrinsicElementPrimitive.fs
alfonsogarciacaro: If you cannot get the list items from the expression you can just spread it. There's no spread operator in Fable AST but you can emit JS (pseudocode): `Emit("...$0", [expr])`

Zaid-Ajaj: Wait, you can spread any arbitrary iterable like that?! 😱 I will need to play with this some more for sure

alfonsogarciacaro:
I've pushed a new Fable.AST version, can you please give it a try? (You need to update Nagareyama too). Also, it should be ok during development but if you're planning to keep the Fable.React dependency, maybe we should move the `ReactComponent` attribute there to avoid user confusion if they see two similar attributes?

Shmew:
I believe the only reason we have the actual `Fable.React` dependency is just so we can make sure things return the same types like `ReactElement` and `IRefValue`, I think we have our own implementation for the entire API at this point. Also wouldn't the plugin function differently since `Fable.React` uses the two list syntax?

Shmew:
A little off-topic but a bit related as I am going to implement this partially as a plugin to clean up the generated code a bit futher... I took another crack at implementing units of measure in place of `ICssUnit` so that you can do things such as `style.marginLeft 10<px>` as opposed to `style.marginLeft (length.px 10)`. I got this all to work, and it properly gives compiler errors if an invalid UoM is provided... but Fable itself (both Fable 2 and Fable 3) can't resolve the overload, while normal F# can. Do you know if there is another way to type check the UoM in the IDE while having Fable properly resolve the overload? Or is this something that just needs to wait for some of the pending items like witnesses/quotations?

```fsharp
type [<Measure>] cm
type [<Measure>] em
type [<Measure>] ex
type [<Measure>] inch
type [<Measure>] mm
type [<Measure>] pc
type [<Measure>] perc
type [<Measure>] pt
type [<Measure>] px
type [<Measure>] rem
type [<Measure>] vh
type [<Measure>] vmin
type [<Measure>] vmax
type [<Measure>] vw

[<Erase>]
module InternalUoM =
type D1 = D1
type D2 = D2
type D3 = D3
type D4 = D4
type D5 = D5
type D6 = D6
type D7 = D7
type D8 = D8
type D9 = D9
type D10 = D10
type D11 = D11
type D12 = D12
type D13 = D13
type D14 = D14

type ToLength = ToLength with
    [<Emit("$1 + 'cm'")>]
    static member (&.) (ToLength, x: int<cm>) = fun D1 _ _ _ _ _ _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'em'")>]
    static member (&.) (ToLength, x: int<em>) = fun D1 D2 _ _ _ _ _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'ex'")>]
    static member (&.) (ToLength, x: int<ex>) = fun D1 D2 D3 _ _ _ _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'inch'")>]
    static member (&.) (ToLength, x: int<inch>) = fun D1 D2 D3 D4 _ _ _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'mm'")>]
    static member (&.) (ToLength, x: int<mm>) = fun D1 D2 D3 D4 D5 _ _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'pc'")>]
    static member (&.) (ToLength, x: int<pc>) = fun D1 D2 D3 D4 D5 D6 _ _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'perc'")>]
    static member (&.) (ToLength, x: int<perc>) = fun D1 D2 D3 D4 D5 D6 D7 _ _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'pt'")>]
    static member (&.) (ToLength, x: int<pt>) = fun D1 D2 D3 D4 D5 D6 D7 D8 _ _ _ _ _ _ _ -> x
    [<Emit("$1 + 'px'")>]
    static member (&.) (ToLength, x: int<px>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 _ _ _ _ _ _ -> x
    [<Emit("$1 + 'rem'")>]
    static member (&.) (ToLength, x: int<rem>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 _ _ _ _ _ -> x
    [<Emit("$1 + 'vh'")>]
    static member (&.) (ToLength, x: int<vh>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 _ _ _ _ -> x
    [<Emit("$1 + 'vmin'")>]
    static member (&.) (ToLength, x: int<vmin>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 D12 _ _ _ -> x
    [<Emit("$1 + 'vmax'")>]
    static member (&.) (ToLength, x: int<vmax>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 D12 D13 _ _ -> x
    [<Emit("$1 + 'vw'")>]
    static member (&.) (ToLength, x: int<vw>) = fun D1 D2 D3 D4 D5 D6 D7 D8 D9 D10 D11 D12 D13 D14 _ -> x

let inline toLen x = unbox<string> (Unchecked.defaultof<ToLength> &. x)
```

Repl link is [here](https://fable.io/repl/#?code=PYBwpgdgBAYghgIwDZgHQGFgCcwFgBQoksiKG2aAUgM4CSEALmFqAUdAMoCe1TAtgQIMu4KAG0APAFkwcagFccAPgC6UAMYD8w0ZJlzFYVVDBadYcdNkLlasAA8hIi3uuHjASwjqAFk91WBrZQfGbOlvo2Rmog6v4ugVHG4Fhx2uGuQdFQIAzxEW7BII7pAZHuajhhZYXZAG5+pQnlwXV8XvmZSWptcCXmBVnGdQDugviSAKJYctEEfMAAJvIoUPRMWBBwSACqwFJQALwEUKdQAwAiAIxHUNcnZ5cATLcXTw+nlwDMr18f5+ELgAWV5A-6XACsrwh4MBADZXnDYaILgB2V6o5EWC4ADleOKxdwAnK8iYTrgAGV5XCnkq43Q53el0l6M67vfBnAEoq4-Nm8ukg-lg-D-QkAFWAABlIABzBg+W6SmUQeWKkYeBX-LlTdoMAAUACIACQ3ADUUAA5JpLYaAJSqbVnXhwBgedQhUwIZhQfUAMlQdt9yrlCoANFB7AAuKBeBgSTRKIOMgBm8mg1ygAH1s7mc-m84WCwWALRKSNO066zVG01QC2W0y2h0qStQF1uj18L0+-2B4PS0M+CPR2OMCSmJO3NMZm5vIsL4uLqBliucrlnasGk3mq0OZuO9cb9sMV3uz18b1YX0BoP6kOq8ORmNxif2Kep9NMu4vC4-JcAcuOariUx5VhIkx6rWu6Wl4vgHq2R4bh257dpeva3gOKpqiOL7jnBPgflAM7fvOf53CCgFUYWIFtlu0H1laoQIW2KFdj2159neD44c+Y7xqEREkZmZE-MCdxQtR1G0UhOoQVBO6MZasQsbJzqnp2F5Xje-b3oOj7Dnxr6xEJX4ib+YkghcUIXAiUkLjJYGWJBNaKQ2KTqKpTlsVpGG6TxT6jsZzDqKZs4-nclkSXcCJokBDnlqBYH0W5Vq5F5YE+Wh2lcVhQ64fxEi5GFpEWRR0W2Xc6K4vFpaJXR8muXW7n2Blx5ZRxOncfpvFBeOxQleZkXldZMVVXceIXCS9mOcljXbs1VpVG1yEaahnW5Xp2GBXh8ZVINc5leJo2VXFNVTUyVLSfVangS5C0wQ0K1ch16GcZhW35UZ44NAdEXkcdNmxdVk0kpSTI3IBs3Hili2Wm0XjPepZ7sW9XV5QZBWvgjEB-aJI1A+N51gzSENMi8S7QxusOPXwfRI6cr05R9AWGX18a9O+ybEWZh3DYDY1naDl1k+yTL-rmVNyfdDENqMDMnijvnvf5PU7YVox40dVmE0LxIi9cc5XL+vJMpRK43f8KAMGOSBeBYDD6ZGtzpggwD2BIvBYF4srlvqOzeD4YDqAA1mAiyoIsYApnAKyOymEis+WAaRna4zsGsjDMFsuz7OM1vnGAvC3I7KpQFcE58EoQA&html=Q&css=Q). In Fable 2 it always just grabs the first method, where Fable 3 it complains that it can't resolve the overloads.

Feliz.CompilerPlugins/ReactComponent.fs
Zaid-Ajaj: @alfonsogarciacaro This might be harder than I thought it would be 😓 I can only transform the call-site but I cannot transform the generated import in the file:
```js
import { counter as Examples_counter } from "./Examples.fs.js" 
```
and inside `Example.fs.js`, the function is exported as PascalCase because of the transformation
```js
export function Counter() { 
 // ... 
}
```
Shouldn't the import be using the transformed `declaration.Name` instead of the default generated one by Fable?

Zaid-Ajaj:
@alfonsogarciacaro The new compiler APIs work great! The only issue right now is what commented above regarding transforming declaration names: when the component becomes upper-case, the import statement stays lower-case which is really really a pity because I would love it if we didn't have to tell the users that components must be upper-case all the time :smile: 

Also don't worry about the attribute coming from `Fable.React`, I don't think anyone is opening both namespaces in one file, there is just no reason to.

Feliz.CompilerPlugins/ReactComponent.fs
alfonsogarciacaro: Hmm, not sure I understand. If you change the `Selector` property in the Import expression, the generated JS import should change as well. Did you try something like this?

```fsharp
| Fable.IdentExpr ident -> Fable.IdentExpr { ident with Name = capitalize ident.Name }
| Fable.Import(info, t, r) -> Fable.Import({ info with Selector = capitalize info.Selector }, t, r)
| _ -> callee
```

alfonsogarciacaro:
I was going to say "please don't go too crazy with plugins and use them for everything now" ;) but it's good we test the limits early on and see how can they be improved before freezing the AST.

Hmm, for the overloads Fable generates a [hash based on the type of the parameters](https://github.com/fable-compiler/Fable/blob/nagareyama/src/Fable.Transforms/OverloadSuffix.fs) but I think we're not getting units of measure into account. Should be possible to fix it? Can you please write a unit test for that and send a quick issue to the Fable repo?

> I believe the only reason we have the actual Fable.React dependency is just so we can make sure things return the same types like ReactElement and IRefValue for interop.

We could put these common elements in a separate library if you want to avoid the full dependency. Not sure if there are still many users mixing the syntaxes though.

Zaid-Ajaj:
> Hmm, not sure I understand. If you change the Selector property in the Import expression, the generated JS import should change as well. Did you try something like this?
> ```fs
> | Fable.IdentExpr ident -> Fable.IdentExpr { ident with Name = capitalize ident.Name }
> | Fable.Import(info, t, r) -> Fable.Import({ info with Selector = capitalize info.Selector }, t, r)
> | _ -> callee
> ```

Thanks a bunch @alfonsogarciacaro, I was not looking in the right place. I tried searching for a _selector_ expression but it was inside the import one. I just implemented it and judging by the generated code, it seems to work just fine :smile: 

> "please don't go too crazy with plugins and use them for everything now"

Speaking of which, can you please have another look into the `[<ReactComponent>]` implementation? 🙏 Especially when it comes to handling the records with `Key` fields.

At this point, I won't be using the other attributes, they seem to have somehow broken what Fable 2.x does and I don't want breaking changes (at least as few as possible) 

Next step is to get a beta version of Feliz out there and re-investigate the fast-refresh work we did before to see if there is more tweaking needed to the compiler plugin

Feliz.CompilerPlugins/ReactComponent.fs
```diff
@@ -0,0 +1,93 @@
...
+        | Fable.Call(callee, info, typeInfo, range) when List.length membArgs = List.length info.Args ->
+            let callee =
+                match callee with
+                | Fable.Expr.IdentExpr ident ->
+                    // capitalize same-file references
+                    Fable.Expr.IdentExpr { ident with Name = AstUtils.capitalize ident.Name }
+                | Fable.Expr.Import(importInfo, fableType, sourceLocation) ->
+                    // capitalize component imports from different modules/files
+                    let selector =
+                        match importInfo.Selector with
```
alfonsogarciacaro, line 24: Ah, forgot this is `Expr`, it should actually be a string. I think I did it like that because in some situations we were adding string literals which are resolved later. But it complicates things, so I've just changed `ImportInfo.Selector` (and Path) to be string. Can you please upgrade to latest Fable.AST? This was the last change before the beta release so we need to be more careful now! :)

alfonsogarciacaro:
That was probably a bad explanation from my side, sorry! Latest implementation looks good! I just left a comment: https://github.com/Zaid-Ajaj/Feliz/pull/266/files#r510635739

Shmew:
> Hmm, for the overloads Fable generates a hash based on the type of the parameters but I think we're not getting units of measure into account. Should be possible to fix it? Can you please write a unit test for that and send a quick issue to the Fable repo?

Sure thing!

> We could put these common elements in a separate library if you want to avoid the full dependency. Not sure if there are still many users mixing the syntaxes though.

Yeah I don't think there's a ton, I'd wager there's a least a few doing so with [Feliz.Plotly](https://github.com/Shmew/Feliz.Plotly). I think this would still be a good idea.

Zaid-Ajaj:
Yolo! let's get this `[<ReactComponent>]` attribute out there. Documentation and docs app rewrite to follow once things become stable in Fable 3. Right now everything is backward-compatible, no need for major version bump (yet)

Feliz v1.16.2 is published as of the latest commit in this PR

alfonsogarciacaro:
Awesome, great work @Zaid-Ajaj! 💪 We should make an announcement so users can start playing with it and we can finish polishing Fable.AST and the plugins before Fable 3 stable release. Does this work already with Fast Refresh?

I'm still worried about the confusion it may happen if we have two versions of the `ReactComponent` plugin for Fable.React and Feliz. For new examples with Fable 3, I'd like to show Feliz as the "recommended" way to write React apps, but there's still a lot of code using the "old" Fable.React bindings and we have to provide support for that. I'm also assuming many apps may move to Feliz incrementally.

A solution could be to share the plugins library between React and Feliz. This should be possible right now because FelizPlugins doesn't have a dependency on Feliz. In the same and as discussed above, we could extract the common Fable.React interfaces used by Feliz so Feliz-only apps don't need to compile the full Fable.React package. Something like:

```
ReactPlugins
ReactTypes
--> Fable.React
--> Feliz
```

Fable.React and Feliz packages are independent but both take the ReactPlugins and ReactTypes dependencies. What do you think?

Zaid-Ajaj:
> We should make an announcement so users can start playing with it and we can finish polishing Fable.AST and the plugins before Fable 3 stable release. Does this work already with Fast Refresh?

In principle, it does work with my very simple counter app. However, I haven't tested this well enough with a larger application to say for sure. The problem is that all the tooling around it is in transition: beta Fable, beta parcel doesn't work but the nightly release does, couldn't make webpack 4 work with fast-refresh, need to figure out how webpack 5 changes things and how to make it work with fast-refresh etc. etc. I think we should not rush things for the announcements before we have a solid story with examples and templates. At least, that is my plan for the time being. I merged this to master because the version is backward-compatible and didn't want to keep rebasing whenever a new PR comes in. 

> Fable.React and Feliz packages are independent but both take the ReactPlugins and ReactTypes dependencies. What do you think?

That sounds like a good idea! If Feliz doesn't depend on Fable.React anymore then they don't need to share the plugin implementation either. I would like to keep `Feliz.CompilerPlugins` here because I am experimenting with plugins that only make sense from Feliz

Shmew:
I'll compile my big work app against this when I can find some time

nojaf:
Hi all, I tried out the latest Feliz (1.16.2) and Fable (3.0.0-nagareyama-beta-005) with [Snowpack ](https://www.snowpack.dev/#create-snowpack-app-(csa))(that uses react-refresh) and can confirm that it works.

alfonsogarciacaro:
That's great @nojaf, thanks a lot for confirming! Would it be possible to share a repo with a simple configuration of Feliz, Fable and Snowpack working together?

nojaf:
@alfonsogarciacaro , sure I pushed the sample to https://github.com/nojaf/snowpack-fable-feliz-sample.
Running `yarn start` after installing all deps should spin up the dev server.
Then changing `SupaHotFire.fs` should refresh everything.

PR #269, opened 2020-10-23
Added min property as DateTime

### LINKED ISSUE #267: Trying to use prop.min with a DateTime but it only allows int or float
Hi,

I tried using prop.min with a DateTime, however, it wasn't possible as it only allows int or float values.
To get around it I did the following which seemed to work:

`Interop.mkAttr "min" minDate` where `minDate = myDate.ToString("yyyy-MM-dd")`

So a possible solution could be to add an overload to the `min` property like so:

```
static member inline min (value: DateTime) = 
  let dateAsString = value.ToString("yyyy-MM-dd")
  Interop.mkAttr "min" dateAsString
```

If you're happy with that solution, I'm happy to create a PR for it. Thanks

CallumVass:
Fix #267


Shmew:
Hey @CallumVass thanks for the contribution!

It looks like this was run through Fantomas or something, as the formatting is radically different on the files you changed. Would you be able to revert those?

For testing you don't need to do anything with the docs application, some of the things that were done for  testing in the docs needs to be moved to our tests project. They all pre-date to me writing bindings for `react-testing-library`.

It looks like `max` has the same missing overload, think you could add it there as well? ❤️ 

Finally, did the two methods I outlined in the issue not work? I tested it in a React repl before I commented and it worked fine:

```javascript
<input value={Date.now()} />

// and

<input value={Date.now().valueOf()} />
```

CallumVass:
Hi @Shmew ,

Sure I will revert changes to other pages. I'm not setting the value, I'm setting the min value, when I tried it in the Example.fs inputs method it was working as expected (I couldn't select dates before today) so I didn't try valueOf

CallumVass:
Hi, updated PR with max value and removed other changes

Shmew:
Yeah `value` was just an example, it should work with `min` and `max` as well.

CallumVass:
No I tried valueOf() and it doesn't seem to work. I can select any date. This is the generated HTML:

    <input type="date" min="1603462163753" max="1603724963753">

According to spec it needs to be in `yyyy-mm-dd` format which is how the PR is currently.

Shmew:
Ah I wasn't understanding that you tried it. Cool, LGTM!

Zaid-Ajaj:
Thanks a lot @CallumVass for the PR ❤️ Feliz v1.15 has been pushed and it includes these new properties

Issue #66, opened 2019-10-25
Solidify Recharts API: Use specialized attribute types for specific elements


Zaid-Ajaj:
Instead of 
```fs
Recharts.lineChart : IReactProperty list -> ReactElement
```
have
```
Recharts.lineChart : ILineChartProperty list -> ReactElement
```
Where `ILineChartProperty` is the specialized property type for the `Recharts.lineChart` element. 

functions within the `lineChart` type will return `ILineChartProperty` and so on and so forth for the rest of the types. 

Of course would be awesome if someone would like to give this a try with PR, let me know if you want to take it!

Shmew:
This is how I do the [Feliz.Plotly](https://github.com/Shmew/Feliz.Plotly) attributes, so there's additional material to take from if needed.

Zaid-Ajaj:
This has been fixed :) thanks @kaashyapan

Issue #58, opened 2019-10-19
More Recharts examples


Zaid-Ajaj:
Help make this project and it's ecosystem better by implementing more example of Recharts, either translating them from the original docs (and filing issues for missing pieces) or creating your own advanced examples that could help out the community

samme78:
I've implemented another one of the Pie charts examples. Mostly to get a feel for the github workflow and learn how to contribute to Open source projects.

Anyways, I'm missing the dominantbaseline prop in Feliz. Its used by the original recharts example (https://jsfiddle.net/alidingling/c9pL8k61/ line 13).

Do you have time to add it? Or could you point me to where and how I should add it?

Zaid-Ajaj:
> I've implemented another one of the Pie charts examples. Mostly to get a feel for the github workflow and learn how to contribute to Open source projects.

Hello @samme78, that is great to hear, I would love to see it added to the docs! 

> Anyways, I'm missing the dominantbaseline prop in Feliz. Its used by the original recharts example (https://jsfiddle.net/alidingling/c9pL8k61/ line 13).

> Do you have time to add it? Or could you point me to where and how I should add it?

I have just added it to the standard properties and published a new version as of 0.65.2 to see how it is implemented, see [this comment](https://github.com/Zaid-Ajaj/Feliz/commit/e366a83e655a7e00e0c52f84c3a3795e99af1592) so next time you know where to look if you want to add it your self. 

If you have any questions or help with how the docs work, please let me know!

PR #270, opened 2020-11-01
Use specialized attribute types for charts

kaashyapan:
Trying to address #66

### DIFF
--- a/Feliz.Recharts/Area.fs
+++ b/Feliz.Recharts/Area.fs
@@ -1,83 +1,82 @@
 namespace Feliz.Recharts
 
 open System
-open Feliz
 open Fable.Core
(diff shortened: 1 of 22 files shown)


Zaid-Ajaj:
Thanks a lot @kaashyapan, this is really awesome! I've been wanting to tackle this issue for a while now :smile: 

I see you also added other types of charts to the main `Recharts` module such as `funnelChart` and `scatterChart`. That's great, now we only need live examples in the docs for them but that is part of #58 :wink:

Zaid-Ajaj:
Published Feliz.Recharts v3.0 that includes these types and enhancements. 

The major version bump is due to breaking changes in the public API, though existing code should work if following the docs and not using generic `IReactProperty` in the input

PR #277, opened 2020-11-19
Added grid-layout options to Styles

Linschlager:
Implemented fully typed and overloaded style options to nicely work with the CSS grid layout in Feliz.
Also enriched the comments (and editor tooltips) with code usage examples for easier usage.

See References here:
[https://css-tricks.com/snippets/css/complete-guide-grid](https://css-tricks.com/snippets/css/complete-guide-grid)
[https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout)

There are a few things that I couldn't implement (yet):
- [grid-template](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template) - I couldn't find a way of properly representing this in F#. It is only a shorthand and no actually missing functionality
- [grid-area](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-area) - The additional overloads as shorthand for `grid-column-end` + `grid-column-start` + `grid-row-end` + `grid-row-start` would have meant 4^2^3 = 4096 overloads for all possible combinations of `string` - `int` - `float` - `IGridSpan`

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -257,6 +257,1430 @@ type style =
     /// Sets the flex grow factor of a flex item main size. It specifies how much of the remaining
     /// space in the flex container should be assigned to the item (the flex grow factor).
     static member inline flexGrow (value: int) = Interop.mkStyle "flexGrow" value
+    /// Sets the width of each individual grid column in pixels.
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-template-columns: 199.5px 99.5px 99.5px;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// gridTemplateColumns: [199.5;99.5;99.5]
+    /// ```
+    static member inline gridTemplateColumns(value: float list) =
+        let addPixels = fun x -> x + "px"
+        Interop.mkStyle "gridTemplateColumns" ((List.map addPixels >> String.concat " ") (unbox<string list> value))
+    /// Sets the width of each individual grid column in pixels.
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-template-columns: 199.5px 99.5px 99.5px;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// gridTemplateColumns: [|199.5;99.5;99.5|]
+    /// ```
+    static member inline gridTemplateColumns(value: float[]) =
+        let addPixels = fun x -> x + "px"
+        Interop.mkStyle "gridTemplateColumns" ((Array.map addPixels >> String.concat " ") (unbox<string[]> value))
+    /// Sets the width of each individual grid column in pixels.
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-template-columns: 100px 200px 100px;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// gridTemplateColumns: [100; 200; 100]
+    /// ```
+    static member inline gridTemplateColumns(value: int list) =
+        let addPixels = fun x -> x + "px"
+        Interop.mkStyle "gridTemplateColumns" ((List.map addPixels >> String.concat " ") (unbox<string list> value))
+    /// Sets the width of each individual grid column in pixels.
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-template-columns: 100px 200px 100px;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// gridTemplateColumns: [|100; 200; 100|]
+    /// ```
+    static member inline gridTemplateColumns(value: int[]) =
+        let addPixels = fun x -> x + "px"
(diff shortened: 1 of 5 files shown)


Shmew:
Hi @Linschlager, thanks for the contribution!

A couple things I noticed:
* grid is in the `Feliz.Styles` namespace, this should probably be in the `style` module or just available under the `Feliz` namespace like `length`, unless there are conflicts. @Zaid-Ajaj thoughts?
* There's a few instances of `sprintf`, we should try to avoid them for performance reasons.
* Missing float overloads for values that input to pixel.

Zaid-Ajaj:
I really like what I am seeing! Thanks a lot @Linschlager for these additions :100: 

I only have a couple of minor comments (along with @Shmew's review) which are the following:
 - `grid` is very generic and used to provide values for `style.gridTemplate*`, `style.gridRow*`, `style.gridColumn*`, I would prefer a more specialized "entry module" like `gridRow`, `gridColumn` and `gridTemplate` which provide the same `span` function and other functions when necessary

 - The docs sometimes are just a bit misleading, using `style.gridGap 1 20` instead of `style.gridGap(1, 20)` and using redundant parentheses. If you could do a quick sweep of these issues in the docs, would be awesome 🙏 

> There are a few things that I couldn't implement (yet): grid-template and grid-area

for things you can't implement in a nice F# API, having at least a string value as input works OK.

Linschlager:
@Zaid-Ajaj Do I understand your first comment correctly; You propose a better readable API with less internal repetition that would look something like this for example:
```f#
style.gridRow.start x
style.gridRow.end x
```

Zaid-Ajaj:
@Linschlager No I meant an API that looks like this
```fs
style.gridTemplateColumns [
  gridTemplate.namedLine "row-1-start"
  gridTemplate.templateWidth (length.fr 1)
]
```
And `gridRow.span`, `gridColumn.span` instead of `grid.span` because I though `grid` is too generic. 

Now to be honest, I haven't used grid styling before so I am not sure if the verbose API helps, what do you think @Linschlager?

Feliz/Styles.fs
```diff
@@ -257,6 +257,1100 @@ type style =
...
+    /// 
+    /// _Shorthand for `rowGap` and `columnGap`_
+    /// 
+    /// **CSS**
+    /// ```css
+    /// gap: 10px 15px;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// style.gap 10 15
```
Zaid-Ajaj, line 693: The docs here should be `style.gap(10, 15)

Feliz/Styles.fs
```diff
@@ -257,6 +257,1100 @@ type style =
...
+    /// 
+    /// _Shorthand for `gridColumnStart` and `gridColumnEnds`_
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-column: 1 / span 2;
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// style.gridColumn (1, (grid.span 2))
```
Zaid-Ajaj, line 1099: The docs here can be simplified `style.gridColumn (1, grid.span 2)`

Feliz/Styles.fs
```diff
@@ -257,6 +257,1100 @@ type style =
...
+    /// 
+    /// _Shorthand for `gridRowStart` and `gridRowEnds`_
+    /// 
+    /// **CSS**
+    /// ```css
+    /// grid-row: row-2 / span "odd-row";
+    /// ```
+    /// **F#**
+    /// ```f#
+    /// style.gridRow ("row-2", (span 2))
```
Zaid-Ajaj, line 1219: Should be `style.gridRow ("row-2", grid.span 2)`

Linschlager:
Ooh, so you mean something like this for example:
```
style.gridTemplate [
  gridTemplate.rows x
  gridTemplate.columns x
  gridTemplate.areas x
]
```

I actually think, there is value to being as close as possible to the CSS API so both people who have already worked with grid and people who just looked at a CSS tutorial and have never heard of it, will get it to work quickly.

I think if the need for a separate API arose, we would need to construct a more thought-through one with the possibilities that F# provides over CSS so it actually provides additional value for example so you need to write less code to achieve the same result and not more as it is now.

Zaid-Ajaj:
@Linschlager That makes sense, let's get this in and see how people use it. We can improve it once we have more feedback :smile:

If you finished with the PR, just let me know to have a last glance before I merge and publish 🚀

Linschlager:
I did a few more doc sweeps and I think I got it right this time. I think this is ready then - I'll mark it as 'Ready for review'

Zaid-Ajaj:
@Linschlager awesome! Merged and published as of Feliz v1.19 🚀

PR #285, opened 2020-11-28
DON'T MERGE: Remove Fable.React dependency

alfonsogarciacaro:
As we've discussed sometimes. This removes Fable.React dependency and instead replaces it with a lightweight Fable.React.Types package that only contains the necessary interfaces.

### DIFF
--- a/Feliz/Html.fs
+++ b/Feliz/Html.fs
@@ -309,8 +309,8 @@ type Html =
     static member inline form xs = Interop.createElement "form" xs
     static member inline form (children: #seq<ReactElement>) = Interop.reactElementWithChildren "form" children
 
-    [<Obsolete("Html.fragment is obsolete, use React.fragment instead. This function will be removed in the coming v1.0 release")>]
-    static member inline fragment xs = Fable.React.Helpers.fragment [] xs
+    // [<Obsolete("Html.fragment is obsolete, use React.fragment instead. This function will be removed in the coming v1.0 release")>]
+    // static member inline fragment xs = Fable.React.Helpers.fragment [] xs
 
     static member inline g xs = Interop.createElement "g" xs
     static member inline g (children: #seq<ReactElement>) = Interop.reactElementWithChildren "g" children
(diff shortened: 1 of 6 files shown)


Feliz/Html.fs
```diff
@@ -309,8 +309,8 @@ type Html =
     static member inline form xs = Interop.createElement "form" xs
     static member inline form (children: #seq<ReactElement>) = Interop.reactElementWithChildren "form" children
 
-    [<Obsolete("Html.fragment is obsolete, use React.fragment instead. This function will be removed in the coming v1.0 release")>]
-    static member inline fragment xs = Fable.React.Helpers.fragment [] xs
+    // [<Obsolete("Html.fragment is obsolete, use React.fragment instead. This function will be removed in the coming v1.0 release")>]
+    // static member inline fragment xs = Fable.React.Helpers.fragment [] xs
```
alfonsogarciacaro, line 313: The comment says the function would be removed in v1.0. As the current version is already far ahead of that I just removed instead of trying to update it.

alfonsogarciacaro:
Unfortunately, dotnet cannot automatically redirect types from one assembly to another so this is causing trouble with the IDEs. If we want to do this we'll have to release as major breaking version and update all packages depending on Fable.React

Zaid-Ajaj:
> If we want to do this we'll have to release as major breaking version and update all packages depending on Fable.React

Wouldn't it be a lot more efficient if I turned `Feliz.CompilerPlugins` into a generic `Fable.ReactCompilerPlugins` package and we reuse both in Fable.React & Feliz?  Breaking changes and having to upgrade every package is no fun for anyone

alfonsogarciacaro:
That'd be nice. But for this PR I was hoping to avoid the extra work and memory needed to compile Fable.React sources for Feliz only applications. But yeah, if it's more pain to update everything then it's not benefitial.

Zaid-Ajaj:
If I only reference the types from Fable.React, do all files still get compiled? Assuming `ReactElement` was in the first file of `Fable.React` (i.e. without dependant files), isn't only that file being compiled?

alfonsogarciacaro:
In Fable 3 all the files are being compiled, we can try to compile only files actually referenced as in Fable 2, but that doesn't worry me too much because the Fable compilation part is quite fast now. What concerns me more is the F# compilation part which takes longer and particularly parsed files remain in memory and are quite heavy.

Anyways, for Fable 3.1 the plan is to try to compile the package sources only once and serialize the inline dependencies, so we don't need to keep the files in memory. If we manage to do that, this shouldn't be too much of an issue.

PR #289, opened 2020-12-09
Fixed padding documentation

Linschlager:
I noticed a small inconsistency on how the padding arguments were named. I updated the naming and documentation for the functions.

Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/padding


Zaid-Ajaj:
Thanks a lot for the PR, this is indeed the correct docs & parameter names. 

Published a new version of Feliz that includes the changes as of v1.22.1 🚀

PR #291, opened 2020-12-11
Add useStateWithUpdater overload

alfonsogarciacaro:
Seems an overload is missing for lazy initialization with `useStateWithUpdater`, is this the right place to put it?


Zaid-Ajaj:
This is the right place, thanks a lot!

PR #295, opened 2020-12-12
Add some missing inlines

bklop:
Just a tiny improvement :). I noticed a call to Html.h1 wasn't being inlined.


Zaid-Ajaj:
Thanks @bklop, merged and published as part of Feliz v1.26 🚀

PR #300, opened 2020-12-22
Add SVG `width` property to Feliz.svg module

Gastove:
Hello!

I was trying to get rolling with Feliz, and I _think_ the `svg` module is missing the `width` property. `width` is [definitely used](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/width), so I figured I'd PR the addition. Alternatively, if I'm misreading or misunderstanding the code: many apologies!

Thanks for the lovely project!

### DIFF
--- a/Feliz/Svg.fs
+++ b/Feliz/Svg.fs
@@ -375,6 +375,24 @@ type svg =
     /// <feMerge>, <feMorphology>, <feOffset>, <feSpecularLighting>, <feTile>, <feTurbulence>, <filter>,
     /// <mask>, <pattern>
     static member inline height (value: int) = Interop.svgAttribute "height" value
+    /// Specifies the width of elements listed here. For all other elements, use the CSS height property.
+    ///
+    /// HTML: <canvas>, <embed>, <iframe>, <img>, <input>, <object>, <video>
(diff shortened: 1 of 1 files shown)


Gastove:
Hrm. `viewBox` also appears to be missing from the `svg` module, but is present in Properties, along with some things concerned with `marker` at least. I'm happy to help pull those over in this PR, or open a new PR, or back off all together. Let me know what works best for y'all!

Zaid-Ajaj:
Thanks a lot for the PR and for pointing out the missing parts @Gastove 🙏 I somehow missed a couple of the really important SVG properties hahah 😅

I merged your PR and published the changes you have added. Also added the missing properties you mentioned: viewBox, viewPort, markerHeight, markerWidth and className. Should be available as of Feliz v1.31 🚀

Gastove:
Hey thanks! Appreciate the snappy, thorough response, the new version, and the awesome project!

Gastove:
Ah, looks like the types aren't lining up -- `viewBox` et al are winding up with a return type of `IReactAttribute`, not `ISvgAttribute`. PR incoming!

PR #301, opened 2020-12-22
Swap `Interop.mkAttr` for `Interop.svgAttribute` in the `svg` module

Gastove:
Otherwise, SVG methods return attributes that can't be used in SVGs!

### DIFF
--- a/Feliz/Svg.fs
+++ b/Feliz/Svg.fs
@@ -137,7 +137,7 @@ type Svg =
 
 type svg =
         /// Specifies a CSS class for this element.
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Thanks a lot! Also I am an idiot 😓

Gastove:
My pleasure! Thanks again for the rad project :D

Zaid-Ajaj:
The pleasure is mine 😄 please upgrade to v1.31.1 to get the fix 🚀

PR #305, opened 2020-12-31
Just tidying up a little on my way through

Gastove:
- The `points` attribute doesn't use its argument, so winds up returning a
  function instead of an `ISvgAttribute`.
- While not required, `xmlns` is nice to have for SVG documents.


Zaid-Ajaj:
Thanks a lot taking the time for open the PR ❤️  and apologies for the delay 🙏 your code made it to Feliz v1.33 and should be available on nuget soon

PR #307, opened 2021-01-07
Fix inputMode capitalization

Linschlager:
The current capitalisation of `inputMode` throws an error in the javascript console, this PR should fix this issue.

image


Zaid-Ajaj:
Thanks a lot! Sorry for the delay, the fix should be included in Feliz v1.33 that will be available soon on nuget

PR #311, opened 2021-01-15
Updated to  webpack v5. Bumped other npm packages.

roboz0r:
Was wanting to learn more about the js side of things as coming from a C#/F# background. I have attempted to update the dependencies and config file to use the latest webpack. Everything seems to be working but would appreciate any additional input.


Zaid-Ajaj:
Hi there @roboz0r, thanks a lot for opening the PR and apologies for the late reply. The PR looks good in general. I will test it out and see if everything is working like it should because the react-refresh plugin was having trouble with webpack v5. Will get back to you as soon as possible 😄

Zaid-Ajaj:
Published as of Feliz template v3.1.0

Issue #309, opened 2021-01-11
Support for pointer-events and touch-action


bzuu:
Hi, it would be nice to have support for the [pointer-events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) and [touch-action](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action) CSS properties. I'd be happy to submit a PR for this. The one thing I am not sure about is what to do with the SVG only values for pointer-events, e.g. `visiblePainted`. Should these be included? Should they somehow be marked as experimental?

Shmew:
Since SVG is separated from the normal properties you can just copy and paste the `pointerEvents` type and then add the two additional lines (and change the attribute call on all of them of course).

bzuu:
@Shmew thanks, I didn't realise the SVG stuff was separate (or even included at all). I was mostly wondering if the properties should be added or not for the HTML side of things as they are marked as experimental for HTML. Would it perhaps be a good idea to include all values for both HTML and SVG and mark them in HTML with the `Experimental`  attribute? Similar to [this](https://github.com/Zaid-Ajaj/Feliz/blob/bdf64d80f5bd3bb1b3821badb4ca0f98f07cb10a/Feliz/Styles.fs#L2207)?

Shmew:
Yeah you can add them in HTML with the attribute (it doesn't do anything, it's just more for us), at least I don't see an issue. Normally I try to avoid APIs that are not fully supported, but in this case every browser supports it, and has for quite some time. @Zaid-Ajaj can give his thoughts, but I suspect he feels the same way.

bzuu:
I am working on the PR for this, but while doing so I it seems like for SVG there is no support for any events as of yet or am I missing something? I can't find any Feliz SVG samples so perhaps I overlooked something.

As far as I can see  `prop.onClick` returns an `IReactProperty`, but the SVG elements expect `ISvgAttribute`. As the `pointerEvents` and `touchAction` properties which I am working on affect the event behaviour, it does perhaps not make a whole lot of sense to implement this for SVG right now?

Shmew:
I believe those are different from the events you're thinking of, they should work for SVG.

bzuu:
[...]

PR #303, opened 2020-12-27
Add box-sizing and resize styles

bzuu:
It seems box-sizing and resize were not yet supported.

PR #313, opened 2021-01-20
Add pointer events html support

bzuu:
Removes incorrect box-sizing inheritance properties introduced in #303 
Adds support for the pointer-events attribute for HTML elements as discussed in #309


Zaid-Ajaj:
Thanks a lot for these @bzuu and apologies for the delay 🙏 merged as part of Feliz v1.33 and should be available soon on nuget

PR #316, opened 2021-02-11
Add pointer event properties

Titaye:
Here are some pointerEvents binding available since react 16.4.0


Shmew:
Looks like some of these were [already added](https://github.com/Zaid-Ajaj/Feliz/blob/master/Feliz/Properties.fs#L1215-L1237). That's why the CI is failing.

Titaye:
Thanks I wasn't up to date.

Shmew:
I wouldn't close the PR out, you have some that are defined here that aren't in the library currently. Or making a new PR with just the new ones works too.

Issue #296, opened 2020-12-13
Documentation/Tests Rewrite


Zaid-Ajaj:
I believe that the new `[<ReactComponent>]` is in a really good shape now and can be safely used as an alternative to the old `React.functionComponent`. 

What is left to be done is rewriting the docs to use Fable 3 and remove all references to `React.functionComponent`, especially the `React Components` tab that explains the various concepts and the common pitfalls page. Then document all the good stuff you can do now with `[<ReactComponent>]` and how to use it. 

Also rewriting the tests and making sure they still work because those are testing `React.functionComponent`.

PR #319, opened 2021-02-23
Addition of radar chart

JamesRandall:
I've added the radar chart and its supporting polar types and an entry into the docs/public area demonstrating its usage to the Recharts project.

I also added the isAnimationActive property onto the bar chart (apologies - this swept up in the same commit).

I've attempted to use the same approach as the rest of the library.

Please do let me know if I've missed anything or got anything wrong and I'd be happy to learn and correct.


Zaid-Ajaj:
Hello @JamesRandall, thanks a lot for this PR, it looks great and follows the structure nicely 😄

Usually I would merge this right away and finish up the small things myself but since you asked, here are a couple of remarks regarding the docs :wink: 

When you add a documentation page in the `public` directory and add this line
```
fsharp:recharts-radar-simpleradarchart
```
above the example the code, the documentation application will try to look for the *live* sample inside the application using the key `recharts-radar-simpleradarchart`  [somewhere in here](https://github.com/Zaid-Ajaj/Feliz/blob/master/docs/App.fs#L359) which you would have a reference to the component. 

The documentation itself won't show up in the docs application unless you add a route to it like we do [here](https://github.com/Zaid-Ajaj/Feliz/blob/master/docs/App.fs#L747) 

You can run the documentation application using `npm run start:nagareyama` if you want to try the app using Fable 3 or just `npm start` if you want to check that it runs with Fable 2.13.x though in #296 I am suggesting we should no longer need to support both versions 

If this doesn't make sense to you or sounds like it is annoying work (it usually is) then just let me know and I will merge the PR and add the rest of the docs. 

I am here for any question you have ✔️

JamesRandall:
I'd like to tidy that up - thanks for the feedback. The docs was the bit I was most uncertain about and next time I'll know how to do this.

I'll get that done now :)

Zaid-Ajaj:
@JamesRandall This looks good to me. Ready to merge or is there something more you want to add? 😄

JamesRandall:
All done and good to merge :)

MangelMaxime:
Hello,

just so you know it seems like @JamesRandall is using the latest version of npm which changed the format of the `package.json`

I don't know if the format is backward compatible or this means that anyone contributing to Feliz will also need to upgrade their npm version to the latest version.

Zaid-Ajaj:
Oh yeah @JamesRandall, I forgot to ask, just as @MangelMaxime said, why is there update in the package-lock.json file? I can't see which packages needed update for this PR 🤔

MangelMaxime:
@Zaid-Ajaj 

It's probably because @JamesRandall is using npm7 and not npm6.

The changes seems backward compatible:

> Changes to the lockfile
> 
> One change to take note of is the new lockfile format, which is backwards compatible with npm 6 users. The lockfile v2 unlocks the ability to do deterministic and reproducible builds to produce a package tree.
> 
> In prior versions, the yarn.lock files were ignored, the npm CLI can now use yarn.lock as the source of package metadata and resolution guidance. If a yarn.lock file is present, then npm will also keep it up-to-date with the contents of the package tree.
> 
> Running npm install with npm 7 in a project with a v1 lockfile will replace that lockfile with the new v2 format. To avoid this, you can run npm install --no-save.

https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile

JamesRandall:
Can confirm their are no package updates.

If it helps I can restore the previous lock file.

Zaid-Ajaj:
> The changes seems backward compatible:

@MangelMaxime thanks a lot for checking, it should be fine if it is backward-compatible :smile: 

@JamesRandall I have just merged and published a new version of Feliz.Recharts as of v3.2 that includes your additions. It should be available as soon as nuget does its magic 🚀

PR #323, opened 2021-03-03
Added scatterchart and tooltip

halcwb:
A lot of Rechart api is missing. I added scatterchart and tooltip and some extra attrs for line and yAxis.

### DIFF
--- a/Feliz.Recharts/Line.fs
+++ b/Feliz.Recharts/Line.fs
@@ -6,6 +6,14 @@ open Fable.Core
 
 [<Erase>]
 type line =
+    static member inline strokeWidth(value : int) = Interop.mkLineAttr "strokeWidth" value
+    /// The source data, in which each element is an object.
+    static member inline data (values: seq<'a>) = Interop.mkLineAttr "data" (Seq.toArray values)
+    /// The source data, in which each element is an object.
+    static member inline data (values: 'a list) = Interop.mkLineAttr "data" (List.toArray values)
+    /// The source data, in which each element is an object.
+    static member inline data (values: 'a array) = Interop.mkLineAttr "data" values
+
     static member inline name(value: string) = Interop.mkLineAttr "name" value
     static member inline dataKey (value: string) = Interop.mkLineAttr "dataKey" value
     static member inline dataKey (f: 'a -> string) = Interop.mkLineAttr "dataKey" f
--- a/Feliz.Recharts/Tooltip.fs
+++ b/Feliz.Recharts/Tooltip.fs
@@ -1,0 +1,23 @@
+﻿namespace Feliz.Recharts
+
+open System
+open Feliz
+open Fable.Core
+open Fable.Core.JsInterop
+
+
+[<Erase>]
+type tooltip =
+    static member inline separator (value: string) = Interop.mkAttr "separator" value
+    static member inline offset (value: int) = Interop.mkAttr "offset" value
+    static member inline filterNull (value: bool) = Interop.mkAttr "filterNull" value
+    static member inline cursor (value: bool) = Interop.mkAttr "cursor" value
+    static member inline cursor(?stroke: string, ?strokeWidth: int) =
+        let cursor = createObj [
+            "stroke" ==> Option.defaultValue "" stroke
+            "strokeWidth" ==> Option.defaultValue 0 strokeWidth
+        ]
+
+        Interop.mkAttr "cursor" cursor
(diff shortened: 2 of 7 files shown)


Feliz.Recharts/Line.fs
```diff
@@ -6,6 +6,14 @@ open Fable.Core
 
 [<Erase>]
 type line =
+    static member inline strokeWidth(value : int) = Interop.mkLineAttr "strokeWidth" value
+    /// The source data, in which each element is an object.
+    static member inline data (values: seq<'a>) = Interop.mkLineAttr "data" (Seq.toArray values)
+    /// The source data, in which each element is an object.
+    static member inline data (values: 'a list) = Interop.mkLineAttr "data" (List.toArray values)
+    /// The source data, in which each element is an object.
+    static member inline data (values: 'a array) = Interop.mkLineAttr "data" values
```
Zaid-Ajaj, line 15: I don't think the `data` attribute it supposed to go on the `line` type 🤔 only `dataKey` and the data goes on the `lineChart` type instead, see [here](https://github.com/Zaid-Ajaj/Feliz/blob/master/Feliz.Recharts/LineChart.fs#L18) Unless I am missing something from the docs at recharts, the `data` attrbute here should be removed

Zaid-Ajaj (changes requested):
Hi Casper, the additions look great! Thanks a lot for the PR 🙏 I have a small comment about adding `data` to `line` (where it already exists on `lineChart`) and I think it should be removed from there. Can you please have a look?

Feliz.Recharts/Line.fs
halcwb: @Zaid-Ajaj I think line also can have its own data:
http://recharts.org/en-US/examples/LineChartHasMultiSeries

`<Line dataKey="value" data={s.data} name={s.name} key={s.name} />`

Zaid-Ajaj: The link points to the latest documentation of recharts which is currently at v2.0.6 so this could be new functionality I haven't seen before in v1.8.6 which is what Feliz.Recharts depends on (1.8.6 >= version > 2.0.0) so if we add this here, we need to either: update Feliz.Recharts required recharts version and testing every sample against  latest 2.0.6 or just go back to older docs to see if the attribute was indeed available in v1.8.6 🤔 the former requires more work to figure whether there are any breaking changes in v2.x but I think that the API hasn't changed that much so that's a good sign

I hope that made sense, dependency management can be tricky 😅

halcwb: Does it help if I tell you that all the added functionality is working in my app? So, the line data should be working as well.

Zaid-Ajaj: Oh yeah absolutely! So if you used Femto to install the binding, then it means the data attribute is compatible. Ok cool, I will try to merge the PR later today and add a sample to docs to make sure everything works as expected. Thanks a lot again ❤️

Zaid-Ajaj:
Hi @halcwb these additions are merged and published to nuget. They should be available in Feliz.Recharts as of v3.3.0 😄

halcwb:
> Hi @halcwb these additions are merged and published to nuget. They should be available in Feliz.Recharts as of v3.3.0

Great, thanks, Would be nice to talk to you again about some projects I am working on.

Feliz.Recharts/Tooltip.fs
```diff
@@ -0,0 +1,23 @@
...
+    static member inline cursor (value: bool) = Interop.mkAttr "cursor" value
+    static member inline cursor(?stroke: string, ?strokeWidth: int) =
+        let cursor = createObj [
+            "stroke" ==> Option.defaultValue "" stroke
+            "strokeWidth" ==> Option.defaultValue 0 strokeWidth
+        ]
+
+        Interop.mkAttr "cursor" cursor
+    static member inline active (value: bool) = Interop.mkAttr "active" value
+    static member inline content (render: ITooltipProperty list -> ReactElement) = Interop.mkAttr "content" render
```
kerams, line 23: `ITooltipProperty list` does not seem right, the [docs](https://recharts.org/en-US/examples/CustomContentOfTooltip) show that the parameter is an object `({ active, payload, label }) =>`.

Zaid-Ajaj: > so what would be the best way of fixing this?

@kerams I haven't been keeping up with changes in recharts but if you say the parameter object has type `({ active, payload, label })` then using an anonymous record should work

PR #324, opened 2021-03-06
Fix svg translate

bzuu:
The SVG translations should not have units as this makes them invalid, causing errors such as:
```
Error: <g> attribute transform: Expected ')', "translate(400px,400px)".
```


Zaid-Ajaj:
Thanks a lot @bzuu I have merged and published Feliz v1.36 and it should be available in a moment 🚀

PR #325, opened 2021-03-08
Property "key" with Guid overload emits wrong attribute name

TWith2Sugars:
No idea why the last line changed, I used github's built in editor.


MangelMaxime:
The last line changed because Github introduced an empty new line in the file. 

That's totally fine.

Zaid-Ajaj:
Thanks @TWith2Sugars this fix is included in v1.37 which should be available on nuget by now 😄

PR #330, opened 2021-03-14
Enable sourcemaps and debugger

roboz0r:
Updates the template to enable source maps per https://github.com/MangelMaxime/fulma-demo/pull/43#issuecomment-756556346
Attempts to use `skipFiles` with the Chrome debugger still hit files in `node_modules` folder. Unsure how to resolve.


MangelMaxime:
Hello,

when generating the files using VSCode Chrome debugger extension in my case I had these lines:

```json
"sourceMapPathOverrides": {
                "webpack:///./~/*": "${workspaceFolder}/node_modules/*",
                "webpack://?:*/*": "${workspaceFolder}/*"
            },
```

Perhaps, `skipFiles` needs to use something similar ?

Zaid-Ajaj:
Added to the Feliz template as of v3.5 🚀

PR #333, opened 2021-03-17
Fix #331

### LINKED ISSUE #331: Nested React Components fails to compile
### Description

If a function tagged with `[<ReactComponent>]` calls another function also tagged with `[<ReactComponent>]` on the same line of code the compiler fails with the following message `EXCEPTION: Cannot find implementation file /`. If the component is bound to a value prior to being passed the compilation is successful. 

### Repro code

From the Felliz template I modified `App.fs` per below

    module App

    open Feliz

    [<ReactComponent>]
    let Comp3 (children:ReactElement) = Html.div [ children ]

    [<ReactComponent>]
    let Comp2 ()  = Html.h1 "Hello world"

    [<ReactComponent>]
    let HelloWorld() = 
        Comp3 (Comp2 ())

If `HelloWorld` is changed to:

    [<ReactComponent>]
    let HelloWorld() = 
        let comp2 = (Comp2 x)
        Comp3 comp2

Then the code is able to compile. 

### Expected and actual results

Expected: The code compiles.
Actual: Fable compiler reports the following `C:/Projects/Feliz Template/src/App.fs(1,1): error EXCEPTION: Cannot find implementation file /`

### Related information

* Fable version: Fable: F# to JS compiler 3.1.10 
* Operating system: Windows 10. Compiling through VS Code powershell terminal. 

    Feliz Version="1.37.0"
    Feliz.Router Version="3.7.0"
    Fable.DateFunctions Version="3.7.0"

alfonsogarciacaro:


Zaid-Ajaj:
Didn't know Christmas was this early this year ❤️ thanks a lot for the fix!

Zaid-Ajaj:
Published as of Feliz.CompilerPlugins v1.4 -> Feliz v1.38

PR #337, opened 2021-03-23
Handle key prop as argument

alfonsogarciacaro:
When using an argument named `key` React will complain that it cannot be accessed. This change avoids the warning by creating an extra prop `$key` in case we need to access it from code.

### DIFF
--- a/Feliz.CompilerPlugins/AstUtils.fs
+++ b/Feliz.CompilerPlugins/AstUtils.fs
@@ -17,6 +17,13 @@ let makeIdent name: Fable.Ident =
       IsMutable = false
       Range = None }
 
+let makeUniqueIdent (name: string) =
+    let hashToString (i: int) =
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj:
Hi @alfonsogarciacaro this looks like a great addition, thanks a lot! It seems that the project no longer compiles according the Appveyor, can you have a look?

alfonsogarciacaro:
This is what happens when you make a PR by copy and pasting before going to bed (or actually already in bed) 😅 I will fix the build.

alfonsogarciacaro:
String interpolation requires FSharp.Core 5.0 and Feliz.CompilerPlugins still reference 4.7.2, so I've removed it :+1:

Zaid-Ajaj:
Published as of
 - CompilerPlugins v1.5 
 - Feliz v1.40

PR #341, opened 2021-03-30
Fix style.overflow.visible

olivercoad:
Fixes `style.overflow.visible` which didn't work due to a typo.


Zaid-Ajaj:
Thanks! Published as of v1.40.1 🚀

PR #344, opened 2021-03-31
[Feliz.Template] Generate CSS with contenthash

OkayX6:
Hi man,

I had regular annoyances where my CSS would not be instantly refreshed in browsers without having to manually delete the cache in Chrome, when deploying in prod :).

Do you mind having this by default for Prod, just like JS scripts?

Thanks


Zaid-Ajaj:
That sounds like a good default for prod. Thanks @OkayX6, will publish a new version of the template later today 👍

PR #347, opened 2021-04-08
Add style types for width

aghoneim92:
I noticed that the styles inherit, unset, etc are only available for height. I added the same types used in height for width.

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -2218,44 +2218,44 @@ module style =
     [<Erase>]
     type minWidth =
         /// Inherits this property from its parent element.
-        static member inline inheritFromParent = Interop.mkStyle "min-width" "inherit"
+        static member inline inheritFromParent = Interop.mkStyle "minWidth" "inherit"
         /// Sets this property to its default value.
-        static member inline initial = Interop.mkStyle "min-width" "initial"
+        static member inline initial = Interop.mkStyle "minWidth" "initial"
         /// Resets this property to its inherited value if it inherits from its parent, and to its initial value if not.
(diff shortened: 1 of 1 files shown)


Feliz/Styles.fs
```diff
@@ -2194,6 +2194,69 @@ module style =
...
+        static member inline maxContent = Interop.mkStyle "width" "max-content"
+
+        /// The intrinsic minimum width.
+        [<Experimental("This is an experimental API that should not be used in production code.")>]
+        static member inline minContent = Interop.mkStyle "width" "min-content"
+
+    [<Erase>]
+    type minWidth =
+        /// Inherits this property from its parent element.
+        static member inline inheritFromParent = Interop.mkStyle "min-width" "inherit"
```
Zaid-Ajaj, line 2221: Shouldn't inline styles follow React's conventions? i.e.
```fs
Interop.mkStyle "minWidth" "inherit"
```
instead of
```fs
Interop.mkStyle "min-width" "inherit"
```

Zaid-Ajaj (changes requested):
Looks good! Just one thing is to make sure we follow React conventions for inline styles which use camelCasing instead of kebab-case

Feliz/Styles.fs
aghoneim92: That's right! Oddly, in `type minHeight` and `type maxHeight` they are specified with `kebab-case` and they work as well 😅

Zaid-Ajaj: I think React will change them into camelCase but will throw warnings in the console when you use it like that

Zaid-Ajaj (approved):
Looks good now!

PR #358, opened 2021-04-28
Fixed casing of dateTime prop

TWith2Sugars:
React complained it was an invalid DOM property


Zaid-Ajaj:
Thanks a lot @TWith2Sugars 🙏 I just published Feliz v1.44 which includes the fix 🚀

PR #361, opened 2021-05-06
Add width and height to Recharts X, Y axes

kerams:
Some more properties that will let me work around https://github.com/recharts/recharts/issues/2027.


kerams:
And by the way, have you ever had the need to change axis colors? There might be som undocumented options https://github.com/recharts/recharts/issues/463. Could be worth adding too?

Zaid-Ajaj:
Thanks @kerams I would like to keep the API limited to the documented options for now

Zaid-Ajaj:
Added as of Feliz.Recharts v3.5 🚀

kerams:
@Zaid-Ajaj, and what about giving the user the ability to set custom attributes (much like `Fable.React.Props.HTMLAttr.Custom`)?

```fsharp
static member inline custom (key: string, value: string) = Interop.mkXAxisAttr key value
```

PR #362, opened 2021-05-06
Add scope property for th Html element

rfrerebe:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#attr-scope


Zaid-Ajaj:
Published as of Feliz v1.45 🚀

PR #368, opened 2021-05-13
Fixed React complaining about onDblClick being wrong

TWith2Sugars:
React was complaining this is wrong.

Also added a corrected version.

Not sure how you want to handle removing wrong/old versions.


Zaid-Ajaj:
I think changing implementation details should be enough without adding an extra API function that is just an alias

Zaid-Ajaj:
Published as of Feliz v1.45 which made `onDblClick` obsolete for users to migrate to `onDoubleClick`

PR #370, opened 2021-05-26
Added overloads for onChange

TWith2Sugars:
I needed to grab an int and float from an html input element with a type of "number" and created two overloads to grab the value directly.

Is this something useful for others?

### DIFF
--- a/Feliz/Interop.fs
+++ b/Feliz/Interop.fs
@@ -78,3 +78,6 @@ module Interop =
         reactElement name (createObj !!properties)
     let inline createSvgElement name (properties: ISvgAttribute list) : ReactElement =
         reactElement name (createObj !!properties)
+
+    [<Emit "typeof $0 === 'number'">]
+    let isTypeofNumber (x: obj) : bool = jsNative
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -1064,10 +1064,18 @@ type prop =
     
     /// Same as `onChange` that takes an event as input but instead let's you deal with the int changed from the `input` element directly
     /// instead of extracting it from the event arguments.
-    static member inline onChange (handler: int -> unit) = Interop.mkAttr "onChange" (fun (ev: Event) -> handler (!!ev.target?value))
+    static member inline onChange (handler: int -> unit) = 
+        Interop.mkAttr "onChange" (fun (ev: Event) -> 
+            if Interop.isTypeofNumber (!!ev.target?value)  
+            then handler (!!ev.target?value)
+        )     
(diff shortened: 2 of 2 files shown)


Zaid-Ajaj:
Hi there @TWith2Sugars these overloads look useful but I don't want them to break once the input elements change their types to something other than `number`/`range` etc. Maybe you could change the implementation such that it only calls the handler when the `!!ev.target?value` is an instance of `Number`? this way `onChange` won't fire if is not a number

thoughts? 🤔

Zaid-Ajaj:
There is actually a downside to this if onChange won't fire unless `type=number` although it makes sense, it might be a confusing to users who only changed the input type but not their handler that accepts `int` or `float`

Zaid-Ajaj:
@TWith2Sugars Sorry I didn't get what you mean by "create an new input model with a number type" 🤔 I think checking for the runtime type should be good enough: i.e. testing with `typeof`
```fs
// into the Interop module maybe?
[<Emit "typeof $0 === 'number'">]
let isTypeofNumber (x: obj) : bool = jsNative

// onChange
Interop.mkAttr "onChange" (fun (ev: Event) -> 
    if Interop.isTypeofNumber (!!ev.target?value)  
    then handler (!!ev.target?value)
) 
```
What do you think?

TWith2Sugars:
Yeah that would be simpler.

I was thinking along the lines of:

`
module prop =
    module input =
        [<Erase>]
        type number =
            static member inline onChange (handler: int -> unit) = Interop.mkAttr "onChange" (fun (ev: Event) -> handler (!!ev.target?value))
            static member inline onChange (handler: float -> unit) = Interop.mkAttr "onChange" (fun (ev: Event) -> handler (!!ev.target?value))
`

So the usage would be

`
Html.input[
    prop.`type`.number
    prop.input.number.onChange(fun _ -> ())
]
`

Zaid-Ajaj:
Yeah let's not do that, it looks inconsistent with what the other `onChange` handlers are doing. Checking that the runtime type is a number should be enough. (though for `int -> unit` handlers it might give weird results unless we do `ceiling` on the value)

TWith2Sugars:
I'll make the change as soon as I can. Want me to stick the isTypeofNumber in to interop then?

Zaid-Ajaj:
> Want me to stick the isTypeofNumber in to interop then?

@TWith2Sugars yes please!

TWith2Sugars:
I've made those changes btw :)

Zaid-Ajaj:
@TWith2Sugars  Awesome! thanks a lot! I will try to merge and publish a new version as soon as possible

Zaid-Ajaj:
Published as of v1.46 🚀

PR #375, opened 2021-06-11
added % options for x and y properties in SVG

TWith2Sugars:
I pretty much copied this from Pie.fs in Feliz.Recharts


Zaid-Ajaj:
Published as of Feliz v1.47 🚀

PR #376, opened 2021-06-11
Added bindings for RadialBarChart and extended label module types

TWith2Sugars:
I used Pie.fs as the basis for this


Feliz.Recharts/RadialBar.fs
```diff
@@ -0,0 +1,38 @@
...
+    static member inline stroke (color: string) = Interop.mkRadialBarAttr "stroke" color
+    static member inline strokeWidth (width: int) = Interop.mkRadialBarAttr "strokeWidth" width
+    static member inline strokeOpacity (opacity: int) = Interop.mkRadialBarAttr "strokeOpacity" opacity
+    static member inline strokeOpacity (opacity: float) = Interop.mkRadialBarAttr "strokeOpacity" opacity
+    static member inline fill (color: string) = Interop.mkRadialBarAttr "fill" color
+    static member inline fillOpacity (value: int) = Interop.mkRadialBarAttr "fillOpacity" value
+    static member inline fillOpacity (value: float) = Interop.mkRadialBarAttr "fillOpacity" value
+
+    static member inline label(value: bool) = Interop.mkRadialBarAttr "label" value
+    static member inline label(?position:Label.Position,?fill:string) =
```
TWith2Sugars, line 33: Unsure if this is a good way to do this, I used https://recharts.org/en-US/examples/SimpleRadialBarChart to get some initial properties.

Feliz.Recharts/RadialBar.fs
```diff
@@ -0,0 +1,38 @@
...
+        let label = createObj [
+          "position" ==> position
+          "fill" ==> Option.defaultValue "" fill
```
Zaid-Ajaj, line 36: I don't think you need to provide a default value for `fill` because in Fable, Option of something translates to something or undefined in JS so it should be good as is:
```fs
createObj [
  "position" ==> position
  "fill" ==> fill
]
```
unless of course one of the options is _required_ in which case a default value would do but here I think it should be fine

Zaid-Ajaj:
Published as of v3.7 🚀

Feliz.Recharts/RadialBar.fs
TWith2Sugars: I think I just forgot to take that out

Zaid-Ajaj: No worries, I refactored it before publishing ;)

PR #378, opened 2021-06-18
add rechart cell onclick binding

christoph-feb22:
Hi @Zaid-Ajaj,

I added a binding for the onclick event for Rechart Cell elements.


Zaid-Ajaj:
Thanks for the PR! published as of Feliz.Recharts v3.8 🚀

PR #379, opened 2021-06-19
Fix referenceArea.fillOpacity key name

kerams:


Zaid-Ajaj:
Thanks for the fix! published as of Feliz.Recharts v3.9 🚀

PR #380, opened 2021-06-23
Fix ChartMouseEvent.activeTooltipIndex

kerams:
Technically a breaking change, but I don't think it would work with the typo right now.


Zaid-Ajaj:
Thanks for the fix! Published as of v3.10

PR #382, opened 2021-07-05
Added a few more SVG properties

TWith2Sugars:


Feliz/Svg.fs
```diff
@@ -800,11 +800,18 @@ type svg =
 
     /// SVG attribute to define the color (or any SVG paint servers like gradients or patterns) used to paint the outline of the shape.
     static member inline stroke (color: string) = Interop.svgAttribute "stroke" color
+    
+    /// SVG attribute to define the pattern of dashes and gaps used to paint the outline of the shape.
+    static member inline strokeDasharray (value: int array) = Interop.svgAttribute "strokeDasharray" value
```
Zaid-Ajaj, line 805: I don't think just using `value` would work. According to the [docs](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray), the value should be a space-separated string made from integers:
```fs
Interop.svgAttribute "strokeDasharray" (List.map string value |> String.concat "")
```

Zaid-Ajaj (changes requested):
Please make sure `strokeDasharray` has the correct format

Feliz/Svg.fs
TWith2Sugars: That's what I thought too but when working with it I passed the array unmodified and it worked fine. I think somewhere React is implicitly converting the array to a string. Playing about with JS string interpolation it seems to match the behaviour I've accidentally stumbled upon.

Cannot seem to find documentation around this behaviour in react. I'll look for a bit longer but if not do you want me to make the changes you suggested?

TWith2Sugars: Looks like it's just a JS thing: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toString

```JS
[1, 2, 3].toString() = "1,2,3"
```

TWith2Sugars: Let me know if you're happy to leave it or want the changes you suggested. I'm happy to do it either way 😊

Zaid-Ajaj:
If it works as is without changes, happy to keep it that way 😄 merged and published as of v1.48 🚀

PR #386, opened 2021-07-22
use valueAsNumber to get the actual number from the input

TWith2Sugars:
This caught me out, the value from an input is always a string.

This uses the [valueAsNumber](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) and checks for NAN if conversion fails.


Zaid-Ajaj:
I thought we were doing some kind of parsing but apparently not, thanks for the PR and sorry for delay 🙏

Zaid-Ajaj:
Merged and included in Feliz v1.50 🚀

TWith2Sugars:
> were doing some kind of parsing but apparently not, thanks for the PR

Don't worry about the delay I'm just happy to contribute :)

PR #387, opened 2021-07-27
Added extra capture options

TWith2Sugars:
This allows the file input type to make use of the devices camera to capture images and videos without the needs to access the camera directly via JS. Pretty much required if you need to take pictures in a web app on iOS.

Details are here: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture


Zaid-Ajaj:
Merged and included in Feliz v1.50 🚀

PR #398, opened 2021-08-23
Add prop.fill

alfonsogarciacaro:
The "fill" SVG attribute seems to be missing. There's only fill for animations.

### DIFF
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -631,6 +631,10 @@ type prop =
     /// Defines the files that will be uploaded when using an input element of the file type.
     static member inline files (value: FileList) = Interop.mkAttr "files" value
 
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @alfonsogarciacaro the `fill` property for SVG elements is [here](https://github.com/Zaid-Ajaj/Feliz/blob/master/Feliz/Svg.fs#L356) is this what you were looking for?

MangelMaxime:
Also, this PR is added a `member` and not an `static member inline`

alfonsogarciacaro:
@MangelMaxime Ups, you're right 😅 

Ah, ok. What confused me is `prop` does contain other SVG-related attributes, including `fillOpacity` and `fill` for SVG `animate` elements. But if those should actually belong to `Svg` type I guess we can close this 👍

PR #401, opened 2021-08-28
WIP: Hot reloading with useElmish

olivercoad:
I've never been able to get hot reloading to work with `useElmish` so I usually just use regular global style Elmish.

However I wanted to try out vite and [Elmish.HMR doesn't currently work with vite](https://github.com/elmish/hmr/issues/29) so I decided to have another look at `useElmish`. This PR makes it work as expected with hot reloading, at least when testing a simple counter with vite+**preact** (which uses prefresh).

From my testing it seems like state in `useState` is preserved between hot reloads but `useRef` is not. The `useEffect` which resets the state and runs cmds also gets executed when doing a hot reload. This change makes it so that the `state` ref is initialised to `childState` so that it is preserved between hot reloads. `useEffect` checks `isFirstRender` before reinitialising state; because isFirstRender is a useRef, it will get reset to true during a hot reload so that it doesn't reset the state.

I tried testing it with **react** as well but had so many other issues with react-refresh and using hooks and fable. I think something to do with having extra `export function ...` in the same js file as the component might be messing up react-refresh somehow. So if anyone can get it working with react let me know :). So that is to say IDK if the behaviour of useRef and useState is the same with react-refresh as it is with prefresh but hopefully it is.

I've marked it as WIP because I need help on making it handle the initial `Cmd` correctly. See, the current behaviour on hot reloads is to reset the state and run inital Cmd (from init). This PR currently changes it so that on hot reloads it doesn't reset the state but **does** still run the initial Cmd which is (probably?) not good.

### DIFF
--- a/Feliz.UseElmish/UseElmish.fs
+++ b/Feliz.UseElmish/UseElmish.fs
@@ -4,50 +4,6 @@ open Feliz
 open Elmish
 open Fable.Core
 
-[<Struct>]
-type internal RingState<'item> =
-    | Writable of wx:'item array * ix:int
-    | ReadWritable of rw:'item array * wix:int * rix:int
-
-type internal RingBuffer<'item>(size) =
-    let doubleSize ix (items: 'item array) =
-        seq { yield! items |> Seq.skip ix
-              yield! items |> Seq.take ix
-              for _ in 0..items.Length do
-                yield Unchecked.defaultof<'item> }
-        |> Array.ofSeq
-
-    let mutable state : 'item RingState =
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @olivercoad, thanks for looking into this. I'll have to look into the current behavior and will let you know of my findings. 

> Sorry for the delayed response, I was on vacation 🌴

Zaid-Ajaj:
I'll close this PR as considered _stale_ and wait for the solution devised by @MangelMaxime

PR #406, opened 2021-09-10
Add missing prop.list

### LINKED ISSUE #405: Data list - input prop.list missing
Hi,

I have recently tried to use the Html.datalist and I found that when using the Html.Input to bind them together there is no 'list' prop on the Input.

Example

```
Html.input [
    prop.type' "text"
    prop.name "example"
    prop.list "exampleList" <-- Error here because list does not exist
]
Html.datalist [
    prop.id "exampleList"
    prop.children [
        Html.option [
            prop.value true
            prop.text "A
        ]
        Html.option [
            prop.value true
            prop.text "B
        ]
    ]
]
```

Is this something that could be added please or point me in the right direction if I have missed something.

Thanks

MangelMaxime:
Fix #405 

Documentation taken from: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input

### DIFF
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -923,6 +923,9 @@ type prop =
     /// No light is projected outside the cone.
     static member inline limitingConeAngle (value: int) = Interop.mkAttr "limitingConeAngle" value
 
+    /// Value of the id attribute of the <c>datalist</c> of autocomplete options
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Thanks a lot @MangelMaxime ❤️ added this as of v1.52

PR #407, opened 2021-09-30
Added support to CSS break after, before and inside properties.

skamlet:
Added the support to the following CSS properties:
* break-after
* break-before
* break-inside

I followed the documentation from Mozilla:
* https://developer.mozilla.org/en-US/docs/Web/CSS/break-after
* https://developer.mozilla.org/en-US/docs/Web/CSS/break-before
* https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside

I tried to adapt the coding/naming style to what was existing, but let me know if there's some changes to be done :smile:


Zaid-Ajaj:
Hi there @skamlet, thanks a lot for taking the time to implement the feature 🙏 

The code looks great! Though I have a proposal to improve upon the usage. The current implementation uses the code as follows:
```fs
prop.style [
  style.breakBefore (breakBeforeStyle.avoidPage)
]
```
However, you can implement it in a different way (same as `display`) to be used as follows:
```fs
prop.style [
  style.breakBefore.avoidPage
]
```
Which is a lot easier to work with. What do you think?

Zaid-Ajaj:
Thanks @skamlet for implementing the changes 🙏 this feature should now be included in Feliz v1.53 🚀

PR #414, opened 2021-10-19
Use Elmish.Program in useElmish

alfonsogarciacaro:
This fixes https://github.com/fable-compiler/fable-promise/issues/24 following @MangelMaxime suggestion to use the Elmish.Program implementation instead of adding a custom one.

I've tested this in a project of mine and also in Fable.Lit and it seems to work. Feliz tests are also passing. The PR is still missing the dependencies, but if you're open to merge this @Zaid-Ajaj I can add them too.

### DIFF
--- a/Feliz.UseElmish/Feliz.UseElmish.fsproj
+++ b/Feliz.UseElmish/Feliz.UseElmish.fsproj
@@ -22,7 +22,6 @@
     </ItemGroup>
     <ItemGroup>
         <PackageReference Update="FSharp.Core" Version="4.7.2" />
-        <PackageReference Include="Fable.Promise" Version="2.0.0" />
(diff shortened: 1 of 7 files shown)


alfonsogarciacaro:
@Zaid-Ajaj Would it be possible to review the PR when you've a moment and merge if everything's ok? Thanks a lot!

Note: if for some reason you cannot maintain this package anymore, we could publish `UseElmish` hook as an independent package from Feliz (only with Fable.React dependency) but in that case it's probably better to deprecate the current package so users are not confused by two competing options.

Zaid-Ajaj:
Hi @alfonsogarciacaro, sorry for the delay 🙏 I have now merged and published this implementation as of Feliz.UseElmish v1.6 🚀

PR #418, opened 2021-11-01
Fix Feliz.key.o

dawedawe:
Feliz.key.o was creating a `p`instead of an `o`


Zaid-Ajaj:
Thanks @dawedawe, merged and published as of Feliz v1.55 🚀

PR #424, opened 2021-11-08
Fix prop.charset.utf8

MangelMaxime:
Hello this PR fix the casing of `prop.charset.utf8`. I missed it in my previous PR

> Invalid DOM property `charset`. Did you mean `charSet`?


Zaid-Ajaj:
Thanks @MangelMaxime 🙏 merged and published as of v1.56

Issue #417, opened 2021-10-25
Text overflow in docs


IltaySaeedi:
The issue has been fixed in my browser, after adding is-10 class to the referenced component.

By looking at the previous lines I think changing line 1181 to the
`prop.className [ Bulma.Tile; Bulma.Is10 ]`
will fix the issue.

https://github.com/Zaid-Ajaj/Feliz/blob/da40e65d5b822d56eb9fbce59e72f239c6f9dcc6/docs/App.fs#L1181

PR #427, opened 2021-11-10
Fix text overflow in docs

IltaySaeedi:
It is a simple fix for https://github.com/Zaid-Ajaj/Feliz/issues/417 . I didn't create PR until now because I had problem with seeing docs pages after running npm command.
overflow


Zaid-Ajaj:
Thanks @IltaySaeedi ❤️ merged and republished the docs

Issue #428, opened 2021-11-19
Feliz.svg missing mask attribute


duncanworthy:
I can see that the Svg type has the mask element mapped, but there doesn't appear to be the mask attribute mapping defined.
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/mask

duncanworthy:
Also just noticed there seems to be a copy-paste error on the mask element definition, as that seems to be mapped to "marker"

Zaid-Ajaj:
Thanks for the PR, this should be resolved now (fix included from #429)

PR #429, opened 2021-11-19
Add mask attribute and fix mapping of mask element

duncanworthy:
Element mappings for mask updated to point to the corresponding react element "mask" where previously it looked like it had been copy-pasted from marker without the mapping being updated.
Also added the corresponding attribute mapping so that the mask element can be applied to other elements by reference - e.g. `svg.mask "url(#mask1)"`
Related issue - #428


Zaid-Ajaj:
Thanks @duncanworthy merged and published as of Feliz v1.57 🚀

PR #434, opened 2021-12-29
Extended style module with more options

Linschlager:
Added some more options to the Styles module

- Completed `justify-x` and `align-x`
- Added float overload to `length.fr`
- Added more overloads to `margin` and `padding` (Fixed margin naming)
- Added single-argument overload to `gap`
- Fixed `outlineOffset`


Zaid-Ajaj:

PR #437, opened 2022-01-11
Extended style module with multiple filters option

Linschlager:
Added an option to have multiple filters applied to the same element without having to resort to `style.custom`.

_The current implementation copies a lot of code from the existing filter type - There may be a more optimal approach without this duplication_


Zaid-Ajaj:

Issue #443, opened 2022-01-31
Incorrect SVG Attributes


rfrerebe:
Some SVG Attributes are wrong (`strokeDasharray` instead of `stroke-dasharray` [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray))
Some SVG Atributes are missing (stroke-dashoffset [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dashoffset) , stroke-linecap  [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap) , stroke-linejoin [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin) , stroke-miterlimit [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit))

Also added a svg.custom member similar to Html props

Zaid-Ajaj:
Fixed as of v1.58

PR #444, opened 2022-01-31
Updated SVG Attribute:

rfrerebe:
- stroke
- custom

Solves #443


Zaid-Ajaj:

PR #456, opened 2022-02-19
Add Escape key & borderSpacing, letterSpacing, overflowWrap properties

panmau:
- Add Escape key to key module
- Add [borderSpacing](https://developer.mozilla.org/en-US/docs/Web/CSS/border-spacing) property
- Add [letterSpacing](https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing) property
- Add [overflowWrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) property

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -2441,6 +2441,24 @@ type style =
             (unbox<string> bottom) + "px " +
             (unbox<string> left) + "px"
         )
+
+    /// Sets the distance between the borders of adjacent table cells
+    static member inline borderSpacing (spacing: int) = Interop.mkStyle "borderSpacing" spacing
+    /// Sets the distance between the borders of adjacent table cells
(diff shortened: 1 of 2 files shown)


Feliz/Styles.fs
```diff
@@ -2441,6 +2441,24 @@ type style =
             (unbox<string> bottom) + "px " +
             (unbox<string> left) + "px"
         )
+
+    /// Sets the distance between the borders of adjacent table cells
+    static member inline borderSpacing (spacing: int) = Interop.mkStyle "borderSpacing" spacing
```
panmau, line 2446: In this case I'm not quite sure whether this works or if I need to do this:
```fsharp
Interop.mkStyle "borderSpacing" ((unbox<string> spacing) + "px )
```

PR #465, opened 2022-03-11
Fixed AppVeyor pipeline

Linschlager:
- In a previous version, the checked-in binary was removed but is still required by the build scripts
    -> Updated scripts to use the dotnet tool version


Zaid-Ajaj:
Thanks a lot! 🙏

PR #468, opened 2022-03-29
ReactMemoComponent

alfonsogarciacaro:
I don't remember if we had this discussion before but it would be very useful to have the option to use `React.memo` with `ReactComponent` attribute. This seems to work, I had to use another type because `new(memo: bool)` constructor conflicted with `new(exportDefault: bool)`. To my surprise F# optional arguments are not compatible with attribute constructors, the other option would be to use .NET optional arguments, as in `[<Optional; DefaultParameterValue(false)>] memo`

If this looks good, it'd be great to have a release so I can use it in my project :)

### DIFF
--- a/Feliz.CompilerPlugins/ReactComponent.fs
+++ b/Feliz.CompilerPlugins/ReactComponent.fs
@@ -9,13 +9,13 @@ open Fable.AST.Fable
 do()
 
 /// <summary>Transforms a function into a React function component. Make sure the function is defined at the module level</summary>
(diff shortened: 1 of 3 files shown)


Zaid-Ajaj:
Hi @alfonsogarciacaro this looks great! Sorry for the delayed response 🙏 merged and published, should be ready in Feliz v1.62 any moment now 😄

alfonsogarciacaro:
Awesome, thank you @Zaid-Ajaj!

Issue #476, opened 2022-04-25
Recharts tooltip properties not correct type


EspenBrun:
![image](https://user-images.githubusercontent.com/22228208/165061849-86cce268-784b-4419-891a-d33c60876725.png)

PR #477, opened 2022-04-25
Recharts tooltip properties :rocket: fixes (#476)

EspenBrun:


Zaid-Ajaj:
Thanks a lot @EspenBrun for the fixes 🙏 merged and added as of Feliz.Recharts v3.12 which will be available soon

PR #478, opened 2022-04-30
Add sourcelink support!

baronfel:
This adds the Dotnet.ReproducibleBuilds package for drop-in sourcelink support! With this change, VS and Ionide can go-to-definition directly to the code that was packaged!

I also centralized some package parameters in a Directory.Build.props, which is a special MSBuild file that applies its properties to all projects beside or below the file.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,10 @@
+<Project>
+    <PropertyGroup>
+        <PackageProjectUrl>https://zaid-ajaj.github.io/Feliz/</PackageProjectUrl>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="all" />
+    </ItemGroup>
+</Project>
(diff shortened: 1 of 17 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+    <PropertyGroup>
+        <PackageProjectUrl>https://zaid-ajaj.github.io/Feliz/</PackageProjectUrl>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
baronfel, line 5: Nuget and other views have better support for these License Expressions rather than a raw License File.

Directory.Build.props
```diff
@@ -0,0 +1,10 @@
...
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="all" />
```
baronfel, line 8: this package reference enables sourcelink automatically, as well as setting a few project properties like the RepositoryUrl.

Zaid-Ajaj:
Thanks @baronfel 🙏 I've merged and published this as of Feliz v1.63

baronfel:
Thanks @Zaid-Ajaj!

PR #492, opened 2022-06-21
Fix react memo import

### LINKED ISSUE #491: Module not found error when using ReactMemoComponent
Using `ReactMemoComponent` causes an error.

```
Module not found: Error: Can't resolve 'React' in '/home/oliver/repos/moto2/src/Client/output'
```

I think it's because the case, the import should be lowercase `react`.

olivercoad:
Fixes #491


alfonsogarciacaro:
My mistake, sorry! 😅  I actually had the same issue in one of my projects. @Zaid-Ajaj Would it be possible to merge this? Thank you!

Zaid-Ajaj:
Hi there @olivercoad, thanks a lot for the fix! It has been merged and published as of Feliz v1.65 🚀

PR #494, opened 2022-07-12
Fix property name of break-before/after/inside

panmau:
The current implementation of breakBefore, breakInside and breakAfter leads to this warning from React:
> Warning: Unsupported style property break-inside. Did you mean breakInside?

This PR solves this by using the camel case name of the property.


Zaid-Ajaj:
Hi there @panmau and thanks a lot for the contribution 🙏 it has been merged and published as of Feliz v1.66 🚀

PR #503, opened 2022-08-12
Add Transform-Origin to Style module

Linschlager:
Implemented a sensible API for transformOrigin that incorporates the entire spec

(Ref: https://developer.mozilla.org/de/docs/Web/CSS/transform-origin)


Zaid-Ajaj:
Thanks a lot @Linschlager for the contribution 🙏 it has been merged and published as of v1.68 of Feliz

PR #506, opened 2022-08-26
Update plugins to Fable 4

alfonsogarciacaro:
Hi @Zaid-Ajaj! We want to release Fable 4 beta soon so users can update their Feliz apps and check there are no regressions. The AST for Fable 4 should be locked by now, so I've updated the plugins to use it. How do you want to go with this? I think the easiest way for now would be to publish an alpha/beta release of Feliz.CompilerPlugins and then either:

1. Publish a prerelease of Feliz that depends on the new plugins
2. Just ask users to add the dependency of the new Feliz.CompilerPlugins. I think this should make Feliz work without a new release, but I need to test it. I'll try to do it using a local feed for the plugins.

Thanks!

### DIFF
--- a/Feliz.CompilerPlugins/AstUtils.fs
+++ b/Feliz.CompilerPlugins/AstUtils.fs
@@ -33,18 +33,21 @@ let makeStrConst (x: string) =
 
 let nullValue = Fable.Expr.Value(Fable.ValueKind.Null(Fable.Type.Any), None)
 
+let makeCallInfo args: Fable.CallInfo = {
+    ThisArg = None
+    Args = args
+    SignatureArgTypes = []
+    GenericArgs = []
+    MemberRef = None
+    Tags = []
+}
+
 let emitJs macro args  =
-    let callInfo: Fable.CallInfo =
-        { ThisArg = None
-          Args = args
-          SignatureArgTypes = []
-          HasSpread = false
-          IsJsConstructor = false
-          CallMemberInfo = None }
+    let callInfo = makeCallInfo args
 
     let emitInfo : Fable.AST.Fable.EmitInfo =
         { Macro = macro
-          IsJsStatement = false
+          IsStatement = false
           CallInfo = callInfo }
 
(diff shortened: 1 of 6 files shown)


Zaid-Ajaj:
Hi @alfonsogarciacaro and thanks a lot for the PR! I've merged it and published pre-release versions of both Feliz and Feliz.CompilerPlugins as of v2.0.0-prerelease-001 

Users can take these for a spin when using Fable v4 😁

alfonsogarciacaro:
Sorry! Forgot to write it here but at the end I tested locally and found an issue in FCS that I've reported here: https://github.com/dotnet/fsharp/issues/13786

We need to fix this before users can test Feliz with Fable 4, but this won't bring changes to the AST, so the plugins shouldn't need republishing.

alfonsogarciacaro:
The issue has been fixed, so users can start updating their Feliz apps 🎉 Please report in Fable repository any regression you may find. Just be sure to install latest Fable tool and Feliz.CompilerPlugins. See commands below, I will also soon publish a beta release and a blog post with more info.

```shell
# In your repo
dotnet tool install --local Fable --version 4.0.0-snake-island-alpha-022

# In your project
dotnet add package Feliz.CompilerPlugins --version 2.0.0-prerelease-001
```

> **Warning**: If you use Thoth.Json make sure you're using v6.

Zaid-Ajaj:
That's awesome @alfonsogarciacaro ❤️ users can also just update Feliz alone to `2.0.0-prerelease-001` and that will bring the latest compiler plugins to the project as well. 

On a side note: do you think https://github.com/fable-compiler/Fable/issues/3110 will introduce another breaking change here?

alfonsogarciacaro:
That will make it easier for users, thanks @Zaid-Ajaj!

Yes, unfortunately the last change to the AST (after that we'll lock it) will require republishing the plugins. Mainly because I just realized I made a mistake and used LibraryImport kind in the makeImport helper 😅  Had I used UserImport, it wouldn't have been affected, because in that case there's no extra info.

PR #507, opened 2022-09-01
Use stable Fable.AST 4

alfonsogarciacaro:
Sorry for this @Zaid-Ajaj but as expected we broke the AST just two days after publishing the new plugins 😅 

Anyways, we've already locked and published Fable.AST 4.0.0 stable so we won't make any modification until the release of Fable 4.

Would it be possible to publish a new version of the plugins with the fix? Thanks and sorry again!

### DIFF
--- a/Feliz.CompilerPlugins/AstUtils.fs
+++ b/Feliz.CompilerPlugins/AstUtils.fs
@@ -70,9 +70,9 @@ let rec flattenList (head: Fable.Expr) (tail: Fable.Expr) =
     ]
 
 let makeImport (selector: string) (path: string) =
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj:
Thanks a lot @alfonsogarciacaro for the update 🙏 I've published new prerelease versions of both Feliz and Feliz.CompilerPlugins v2.0.0-prerelease-002 🚀 although everything is stable, I will wait for Fable to become a stable v4 before publishing the `2.0.0` version

alfonsogarciacaro:
Sounds good, thank you @Zaid-Ajaj!

Issue #508, opened 2022-09-07
Extract ReactElement to another library and replace Fable.React dependency


alfonsogarciacaro:
This is something we discussed in the past, I tried to do it in #285, but we didn't merge at the end because moving interfaces to another .dll broke .NET builds. If you're planning to release Feliz 2 @Zaid-Ajaj, maybe this can be a good opportunity to try it again. The idea is:

1. Extract Fable.React types (at a minimum `ReactElement`) to another library, say Fable.React.Types, and publish it independently.
2. Publish a new Fable.React version that depends on this library, but probably after that we can deprecate it.
3. Make Feliz 2 depend on Fable.React.Types but no Fable.React. This will still make Feliz elements compatible with Fable.React, but it will avoid the dependency when not needed.
4. Other libraries that basically only need ReactElement can do the same, like elmish-react and the upcoming Feliz.JSX.React.

If this is OK with you I can make the Fable.React split and send a new PR 👍

Zaid-Ajaj:
That sounds great @alfonsogarciacaro 🙏 Feliz v2 sounds like a good candidate for this!

PR #509, opened 2022-09-08
Use Fable.React.Types dependency

alfonsogarciacaro:
@Zaid-Ajaj This is the change discussed in #508 (PR also moves some XML comments to avoid warnings during build). Tests seem to be passing. I wasn't expecting docs to work because of the Fable.Elmish.React dependency but a bit surprisingly they do build. Although I do get a runtime error when trying to run them in a local server. I will send a similar PR to Fable.Elmish.React and try again.

**Note**: I've also added the README file to the packages and the Fable logo. But of course, feel free to discard that if you don't want to use the same README for all packages or if you want to use a different logo.


build/paket.references
```diff
@@ -0,0 +1 @@
+Fable.React.Types
```
Zaid-Ajaj, line 1: I don't think you need Fable.React.Types in the build project, right? It is only for publishing packages and other build automation stuff

build/Build.fsproj
```diff
@@ -1,19 +1,17 @@
...
     <Compile Include="Program.fs" />
   </ItemGroup>
-
   <ItemGroup>
     <PackageReference Include="Fake.Core.Environment" Version="5.20.4" />
     <PackageReference Include="Fake.Core.Target" Version="5.20.4" />
   </ItemGroup>
-
-</Project>
+  <Import Project="..\.paket\Paket.Restore.targets" />
```
Zaid-Ajaj, line 16: Same here, build project doesn't depend on paket and doesn't need to share dependencies with other projects

paket.lock
```diff
@@ -39,10 +43,14 @@ NUGET
     Fable.Promise (3.1)
       Fable.Core (>= 3.1.5) - restriction: >= netstandard2.0
       FSharp.Core (>= 4.7.2) - restriction: >= netstandard2.0
-    Fable.React (7.4)
+    Fable.React (7.4) - restriction: >= netstandard2.0
       Fable.Browser.Dom (>= 2.0.1) - restriction: >= netstandard2.0
       Fable.Core (>= 3.1.5) - restriction: >= netstandard2.0
       FSharp.Core (>= 4.7.2) - restriction: >= netstandard2.0
+    Fable.React.Types (18.0.0-beta-001)
```
Zaid-Ajaj, line 50: Why does this need to be `beta`?

Feliz/React.fs
```diff
@@ -62,16 +62,24 @@ type React =
...
 
     /// The `React.fragment` component lets you return multiple elements in your `render()` method without creating an additional DOM element.
-    static member inline fragment xs = Fable.React.Helpers.fragment [] xs
+    static member inline fragment xs =
+        Fable.React.ReactBindings.React.createElement(Fable.React.ReactBindings.React.Fragment, obj(), xs)
 
     /// The `React.fragment` component lets you return multiple elements in your `render()` method without creating an additional DOM element.
-    static member inline keyedFragment(key: int, xs) = Fable.React.Helpers.fragment [ !!("key", key) ] xs
+    static member inline keyedFragment(key: int, xs) = // Fable.React.Helpers.fragment [ !!("key", key) ] xs
+        Fable.React.ReactBindings.React.createElement(Fable.React.ReactBindings.React.Fragment, createObj ["key" ==> key], xs)
```
Zaid-Ajaj, line 70: No need to convert `xs` to an array here?

Feliz/Html.fs
```diff
@@ -433,12 +433,12 @@ type Html =
     static member inline kbd (value: string) = Interop.reactElementWithChild "kbd" value
     static member inline kbd (children: #seq<ReactElement>) = Interop.reactElementWithChildren "kbd" children
 
-    [<Obsolete("Html.keyedFragment is obsolete, use React.keyedFragment instead. This function will be removed in the coming v1.0 release")>]
-    static member inline keyedFragment (key: int, xs) = Fable.React.Helpers.fragment [ !!("key", key) ] xs
-    [<Obsolete("Html.keyedFragment is obsolete, use React.keyedFragment instead. This function will be removed in the coming v1.0 release")>]
-    static member inline keyedFragment (key: string, xs) = Fable.React.Helpers.fragment [ !!("key", key) ] xs
-    [<Obsolete("Html.keyedFragment is obsolete, use React.keyedFragment instead. This function will be removed in the coming v1.0 release")>]
-    static member inline keyedFragment (key: System.Guid, xs) = Fable.React.Helpers.fragment [ !!("key", string key) ] xs
+    // [<Obsolete("Html.keyedFragment is obsolete, use React.keyedFragment instead. This function will be removed in the coming v1.0 release")>]
```
Zaid-Ajaj, line 436: We can probably remove these entirely in v2 because they have been obsolete for a long time

Feliz/React.fs
alfonsogarciacaro: The children argument of `ReactBindings.React.createElement` is decorated with `ParamList` attribute (alias for `ParamSeq`), this is similar to `ParamArray` but it allows for any `seq` type because in JS you can spread any iterable not only arrays. I added it in Fable 2 specifically for React.createElement so we didn't have to do the runtime array conversion just to use lists in F#.

But if you mean `React.Children.toArray` to avoid the React warning about the keys, I'm not sure, I'd have to check. Although in principle this will generate the same code as before when calling `Fable.React.Helpers.fragment`.

build/Build.fsproj
alfonsogarciacaro: Hmm, I think Paket did this automatically when running `dotnet paket install`. I will undo the changes 👍

paket.lock
alfonsogarciacaro: I always push a beta version first just in case 😅 But you're right, a stable version should be fine as we've been using the bindings for a while now. I will update it 👍

alfonsogarciacaro (commented):
Thanks a lot for reviewing @Zaid-Ajaj! I've implemented your suggestions 👍

Zaid-Ajaj:
Merged and published as of Published Feliz v2.0.0-prerelease-003 🚀

PR #512, opened 2022-09-14
Fix #510: Inject React import for components

### LINKED ISSUE #510: Add `import React from "react"` automatically when the file contains a `[<ReactComponent>]`
In order for React fast refresh to works, the following import  `import React from "react"` needs to be in the file.

Demo:

https://user-images.githubusercontent.com/4760796/189533763-097ee58f-8efc-438a-9b07-be3ed2c2b347.mp4

1. When first editing the file see how the page does a full reload
2. When I edit after adding `import React, { createElement } from "react"` the file is now using hot reload.

One workaround is to add 

```fs
open Fable.Core.JsInterop

// Workaround to have React-refresh working
// I need to open an issue on react-refresh to see if they can improve the detection
emitJsStatement () "import React from \"react\""
```

at the top of each file but this is easy to forget and error prone. Especially, because the user don't need to add it on top of each file, but just the one which export React components.

@alfonsogarciacaro Is it possible to use the `FelizComponent` plugin to auto inject the line `import React from "react"` in the file?

I don't know if having several time the same import has any implication in JavaScript. But, we probably don't want to inject it at each `FelizComponent` occurrence but only once per file.

alfonsogarciacaro:
This injects `import React as "react"` in all files declaring a React component as discussed in #510.

Please note this requires the very latest Fable 4 pre-release, still unpublished 😅 

@MangelMaxime Imports may become a bit overcrowded, for example when compiling Feliz docs I get (among other imports):

```js
import { createElement, memo } from "react";
import React from "react";
import * as react from "react";
```

A solution could be to make Feliz helpers import `* as React`, but this would require upgrading more things for the users. So if there's no issue in the duplicated imports (default `React` and namespace `* as react`) we can leave it as is.

### DIFF
--- a/Feliz.CompilerPlugins/AstUtils.fs
+++ b/Feliz.CompilerPlugins/AstUtils.fs
@@ -114,6 +114,10 @@ let recordHasField name (compiler: PluginHelper) (fableType: Fable.Type) =
     | _ ->
         false
 
+let memberName = function
+    | Fable.MemberRef(_,m) -> m.CompiledName
+    | Fable.GeneratedMemberRef m -> m.Info.Name
+
 let makeCall callee args =
     Fable.Call(callee, makeCallInfo args, Fable.Any, None)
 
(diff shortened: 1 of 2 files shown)


MangelMaxime:
@alfonsogarciacaro 

> A solution could be to make Feliz helpers import `* as React`, but this would require upgrading more things for the users.

If we make the change in Feliz, I think this is will not be a breaking change right? So, users will naturally upgrade to it over time.

> Imports may become a bit overcrowded, for example when compiling Feliz docs I get (among other imports):
> So if there's no issue in the duplicated imports (default `React` and namespace `* as react`) we can leave it as is.

I don't think having duplicated is having a real impact. Perhaps, just a bit of overload because JavaScript needs to process 2 imports instead of one.

Ideally, if what I said above about not the transition being not a breaking change it would be ideal to do it. In general, I try to keep the generated JavaScript on par with what native JavaScript use in general.

Zaid-Ajaj:
Published Feliz.CompilerPlugins v2.0.0-prerelease-003 which includes this change, ready to be tested 😄

PR #515, opened 2022-10-05
UseElmish with useSyncExternalStore

### LINKED ISSUE #481: UseElmish with Fast Refresh
Hi @Zaid-Ajaj,

this is related to WIP PR #423. I usually cannot make UseElmish to work with Fast Refresh, even when following all the rules like PascalCase + making init + update functions private. However, I tried to use @alfonsogarciacaro version and got some minor improvement:

useElmish

As you can see, Alfonso's version survives the first reload, and on the second change, it resets to the initial state again.

Do you have any idea what could be done on the top of Alfonso's PR to survive every refresh (a.k.a. make it working with Fast Refresh)? Thanks for any help! 🙏

alfonsogarciacaro:
Fix #481

This PR updates UseElmish to use `useSyncExternalStore`, which seems to be [the recommended hook for this use case](https://beta.reactjs.org/learn/you-might-not-need-an-effect#subscribing-to-an-external-store) in React 18. The code is greatly simplified, we don't need the ElmishObservable or ResumableState anymore.

It took a bit of trial-and-error to use the hook correctly but [I've tested it](https://github.com/fable-compiler/Feliz.JSX/commit/ecd7c000b24199516d87c2dd1f954d67826c9ca9) and it's working fine with hot reloading.

The PR also upgrades the code to Elmish 4.

~~One breaking change is that I removed the `dependencies` argument. This is because the code now only uses two React hooks: useState and useSyncExternalStore, and neither has a `dependencies` argument. To support dependencies we would have to add `useEffect` or our own logic complicating things. It's already tricky to get the life cycle right and Elmish 4 also introduces termination capabilities, so I'd prefer not to add extra logic to restart the state. If you need to reset the model if a dependency changes, we can create a specific message for this and handling it in the `update` function.~~

@MangelMaxime @Dzoukr

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "fable": {
-      "version": "3.7.11",
+      "version": "4.0.0-theta-011",
       "commands": [
         "fable"
       ]
--- a/Feliz.ElmishComponents/Feliz.ElmishComponents.fsproj
+++ b/Feliz.ElmishComponents/Feliz.ElmishComponents.fsproj
@@ -14,6 +14,7 @@
     </ItemGroup>
     <ItemGroup>
         <ProjectReference Include="..\Feliz.UseElmish\Feliz.UseElmish.fsproj" />
+        <ProjectReference Include="..\Feliz\Feliz.fsproj" />
     </ItemGroup>
     <ItemGroup>
         <Content Include="*.fsproj; *.fs; *.js;" PackagePath="fable\" />
--- a/Feliz.UseElmish/Feliz.UseElmish.fsproj
+++ b/Feliz.UseElmish/Feliz.UseElmish.fsproj
@@ -19,6 +19,6 @@
     </ItemGroup>
     <ItemGroup>
         <PackageReference Update="FSharp.Core" Version="4.7.2" />
-        <PackageReference Include="Fable.Elmish" Version="3.0.6" />
+        <PackageReference Include="Fable.Elmish" Version="4.0.0-beta-5" />
     </ItemGroup>
 </Project>
--- a/Feliz.UseElmish/Feliz.UseElmish.fsproj
+++ b/Feliz.UseElmish/Feliz.UseElmish.fsproj
@@ -12,7 +12,7 @@
         <Compile Include="UseElmish.fs" />
     </ItemGroup>
     <ItemGroup>
-        <ProjectReference Include="..\Feliz\Feliz.fsproj" />
+        <ProjectReference Include="..\Feliz.CompilerPlugins\Feliz.CompilerPlugins.fsproj" />
     </ItemGroup>
     <ItemGroup>
         <Content Include="*.fsproj; *.fs; *.js;" Exclude="**\*.fs.js" PackagePath="fable\" />
--- a/Feliz.UseElmish/Feliz.UseElmish.fsproj
+++ b/Feliz.UseElmish/Feliz.UseElmish.fsproj
@@ -8,6 +8,11 @@
         <TargetFramework>netstandard2.0</TargetFramework>
         <PackageReleaseNotes>Use and accept a full Elmish program instead of re-implementing it</PackageReleaseNotes>
     </PropertyGroup>
+    <PropertyGroup>
+        <NpmDependencies>
+        <NpmPackage Name="use-sync-external-store" Version="&gt;= 1.0.0 &lt; 2.0.0" ResolutionStrategy="Max" />
+        </NpmDependencies>
+    </PropertyGroup>
     <ItemGroup>
         <Compile Include="UseElmish.fs" />
     </ItemGroup>
--- a/Feliz.UseElmish/Feliz.UseElmish.fsproj
+++ b/Feliz.UseElmish/Feliz.UseElmish.fsproj
@@ -12,9 +12,6 @@
         <Compile Include="UseElmish.fs" />
     </ItemGroup>
     <ItemGroup>
-        <ProjectReference Include="..\Feliz.CompilerPlugins\Feliz.CompilerPlugins.fsproj" />
-    </ItemGroup>
-    <ItemGroup>
         <Content Include="*.fsproj; *.fs; *.js;" Exclude="**\*.fs.js" PackagePath="fable\" />
     </ItemGroup>
     <ItemGroup>
--- a/Feliz.UseElmish/UseElmish.fs
+++ b/Feliz.UseElmish/UseElmish.fs
@@ -12,7 +12,7 @@ module private Util =
     [<ImportMember("react")>]
     let useState(init: unit -> 'State): 'State * ('State -> unit) = jsNative
 
-    type ElmishState<'Arg, 'Model, 'Msg when 'Arg : equality>(program: unit -> Program<'Arg, 'Model, 'Msg, unit>, arg: 'Arg, dependencies: obj[]) =
+    type ElmishState<'Arg, 'Model, 'Msg when 'Arg : equality>(program: unit -> Program<'Arg, 'Model, 'Msg, unit>, arg: 'Arg, dependencies: obj[] option) =
         // let guid = System.Guid.NewGuid()
         // do printfn "Creating %O..." guid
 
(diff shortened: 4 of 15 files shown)


MangelMaxime:
Thank you @alfonsogarciacaro for looking into this.

I didn't review the code yet, but have a question.

> If you need to reset the model if a dependency changes, we can create a specific message for this and handling it in the update function.

I don't understand how this will work.

In Fable.Packages, we have one Elmish components which has a dependency over `props.UrlParameters` making it possible to adapt to changes of the URLs.

```fs
    [<ReactComponent>]
    static member SearchForm(props: SearchFormProps) =
        let model, dispatch =
            React.useElmish (
                init props.UrlParameters props.OnSearch,
                update props.OnSearch,
                [|
                    box props.UrlParameters
                |]
            )

        Html.div [
            Components.SearchInputField model dispatch
            Components.Filters(model, dispatch)
        ]
```

https://github.com/fable-compiler/packages/blob/232b5ec1322a30e16ce2bcddc37cff9f007fc201/src/Components/SearchForm.fs#L462-L476

How would this work in this case?

alfonsogarciacaro:
Sorry @Zaid-Ajaj, I did a force push and this seems to confuse Github notifications. This is ready to merge when you have a moment to look at (but please check comment above).

BTW, would it be possible to make @MangelMaxime and I also owners of the Feliz.UseElmish Nuget package so we can make new releases if necessary?

alfonsogarciacaro:
Actually, I just realized if I decorate the type with `CompiledName("useReact")` I can get the same effect as when using the `Hook` attribute. Anyways, in my tests there didn't seem to be a difference between prefixing with `use-` or not, maybe that has been solved already?

Also, when using JSX.Component it may happen there is no `import React` in the file at the end. But I got hot reload working anyways. @MangelMaxime where did you see that `import React` is necessary? I've only seen it for the case of compiling JSX, but it seems this is not necessary with the [new JSX runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).

This is the summary of the things I've noticed when using `@vitejs/plugin-react` for Fast Refresh:

- Having an `import React` clause doesn't seem to be necessary
- Using `export default` doesn't seem to be necessary
- However, there must be **only one export** for Fast Refresh to work. There can be other components in the file (and they will trigger fast refresh too) but they must be private.

> About the 3rd point, one must be careful when using a static class in order to have optional arguments. In that case the class must be decorated with `Erase` so the code is not output adding an extra export.

Also, given that `UseElmish` doesn't have a Feliz dependency anymore, would it make sense to host it in its own repo and have more maintainers for the package? Should we change the namespace in that case (e.g. `Fable.React.UseElmish`)?

alfonsogarciacaro:
@Zaid-Ajaj `npm test` is failing, but this is still running with Fable 2. `npm run test:nagareyama` works for me locally. Do we still want to maintain the tests for Fable 2 or can we skip them? (BTW, I'm also surprised the tests for Fable 3 were working even after updating the plugin)

Zaid-Ajaj:
Hi @alfonsogarciacaro sorry for the late reply here. I will take a closer look at the PR as soon as possible 🤞

alfonsogarciacaro:
Thanks @Zaid-Ajaj! I made a change in [this commit](https://github.com/Zaid-Ajaj/Feliz/pull/515/commits/1463415715c632b6ae31bed637900b5f88514c7d) that fixes the Fable 2 tests, but it exposes a `useElmish` function besides the `React` overloads. I don't have a strong opinion on this, it looks like prefixing with `use` if not necessary now, but if you want we can add back the `Hook` plugin.

MangelMaxime:
> @MangelMaxime where did you see that `import React` is necessary?

I it based on reading the code of react-refresh, and also by experimenting.

I did the experimentation by manually editing the JavaScript, F# files, etc.

It is important to note, that if you are using JSX syntax then Vite, Babel, and other plugin are mostly likely to add code for you.

For example, `vite/plugin-react` is going to prepend the generated code with `import React from 'react';`

https://github.com/vitejs/vite/blob/e976b065628ca451cd07bf5472427462d64bc56c/packages/plugin-react/src/index.ts#L98

https://github.com/vitejs/vite/blob/e976b065628ca451cd07bf5472427462d64bc56c/packages/plugin-react/src/index.ts#L283-L293

So, I fell like you think you are not adding `import React from 'react';` in your F# code , but because you are using JSX it will still be present in the output.

> However, there must be only one export for Fast Refresh to work. There can be other components in the file (and they will trigger fast refresh too) but they must be private.

I have to disagree, with this one. You can have multiple components exported in the same file and Fast refresh is still going to work. Fable.Packages does it for example.

```fs
module Fable.Packages.Components.AnErrorOccured

open Fable.Core.JsInterop
open Feliz
open Feliz.Bulma
open Fable.Packages

// Workaround to have React-refresh working
// I need to open an issue on react-refresh to see if they can improve the detection
emitJsStatement () "import React from \"react\""

type Components with

    [<ReactComponent>]
    static member AnErrorOccured
        (
            title: string,
            subtitle: string,
            details: string
        ) =
        // ...

    [<ReactComponent>]
    static member AnErrorOccured(title: string, subtitle: string) =
        // ...
```

I also made a sandbox, using pure JavaScript:

[Edit React Fast Refresh Test (forked)](https://codesandbox.io/s/react-fast-refresh-test-forked-3z9xng?fontsize=14&hidenavigation=1&theme=dark)

https://user-images.githubusercontent.com/4760796/197342224-5726635a-34a5-4fe2-ba70-5c2686f9a0fa.mov

MangelMaxime:
@alfonsogarciacaro 

Do you think it would be possible to use [use-sync-external-store](https://www.npmjs.com/package/use-sync-external-store) NPM package?

This package allow to have `use-sync-external-store` working for **any** version of react.

The drawback is that users will need to install `react`, `react-dom` **and** `use-sync-external-store` NPM packages. But this is kind of the recommanded way of doing it.

Zaid-Ajaj:
> However, there must be only one export for Fast Refresh to work. There can be other components in the file (and they will trigger fast refresh too) but they must be private.

Last time I checked, there could be multiple exports for React components in one file, so this should be no problem AFAIK

> Do you think it would be possible to use [use-sync-external-store](https://www.npmjs.com/package/use-sync-external-store) NPM package?

It would be great to do this @alfonsogarciacaro! People who use Feliz < 2.0.0 can still use the updated Feliz.UseElmish without having to upgrade everything at once 🙏 could you please have a look at that?

alfonsogarciacaro:
Using `use-sync-external-store` makes sense, we had a similar discussion here: https://github.com/alfonsogarciacaro/fable-react-sample/issues/1#issuecomment-1286542980

In any case, please note that:
- this avoids having to update to React 18 (which is important)
- has no influence on Feliz/Fable.React versions (as discussed above, if we remove the `Hook` attribute we don't even need Feliz nor Fable.React as dependencies).
- still requires updating to Elmish 4 (which may have an impact on devs using "global" Elmish and UseElmish at the same time)

My only concern is that I have already released Fable.React.UseElmish and I'm reluctant to push a new version with a new npm dependency. But maybe we could publish Feliz.UseElmish with `use-sync-external-store` dependency so users who don't want to update to React 18 yet can use it.

As an aside, the React import may has to do with the [new JSX transform in React 17](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#whats-a-jsx-transform), the linked web says this avoids the React import requirement. And as an aside x2, this makes me wonder if we should use `jsx` in the ReactComponent plugin instead of `React.createElement` as the former brings [some performance improvements](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#motivation).

MangelMaxime:
@alfonsogarciacaro I am not sure what `Fable.React.UseElmish` is? 

Is it a fork/port of the code from you PR ? Or is it something else?

Using the new JSX transform make sense I guess.

> My only concern is that I have already released Fable.React.UseElmish and I'm reluctant to push a new version with a new npm dependency.

Make a major bump as it is a breaking changes.

If the package is missing, the bundler will complain about it and user will be inform.

You could also, in DEBUG add a check + custom exception explaining the situation but I think the bundler should be enough.

alfonsogarciacaro:
@MangelMaxime @Zaid-Ajaj Sorry for taking so long. As discussed, I'm using the polyfill now so this is compatible with React 17.

Appveyor was installing dotnet 7 which causes issues with Fable (we have fixed them for Fable 4) so I pinned the sdk to net6.0. Also the tests for UseElmish were failing when running on Fable 2. Not sure why but it's probably not worth to spend time investigating so for now I just set them to run only on Fable 3/4.

PR #537, opened 2022-12-29
Add missing Recharts Components and properties

sydsutton:
I added ErrorBar, Funnel, Legend, and Text, along with some missing properties along the way. Please check the 'onMouseEnter', 'onMouseLeave', etc... properties as I copied them from other components and don't know if they 100% fit with the components I added them to. 

Thanks!


sydsutton:
I'm not sure why it's not passing the "AppVeyor" checks. After fixing the build errors in my last commit, it builds on my machine just fine.

Feliz.Recharts/Text.fs
```diff
@@ -0,0 +1,27 @@
+namespace Feliz.Recharts
+
+open System
+open Feliz
+open Fable.Core
+open Fable.Core.JsInterop
+
+[<Erase>]
+type text =
```
Zaid-Ajaj, line 9: Here you added attributes of type `ITextProperty` but where are these used?

Feliz.Recharts/Legend.fs
```diff
@@ -0,0 +1,94 @@
...
+    static member inline margin (?top: int, ?left: int, ?right: int, ?bottom: int) = 
+        let margin = 
+            createObj [ 
+                "top" ==> Option.defaultValue 0 top
+                "left" ==> Option.defaultValue 0 left
+                "right" ==> Option.defaultValue 0 right
+                "bottom" ==> Option.defaultValue 0 bottom ]
+
+        Interop.mkLegendAttr "margin" margin
+    static member inline formatter (f: string -> 'a -> int) = Interop.mkLegendAttr "formatter" (System.Func<_,_,_> f)
```
Zaid-Ajaj, line 24: Why is the lambda type `string -> 'a -> int` what is `formatter` supposed to do?

Zaid-Ajaj (changes requested):
Looks good in general. Have a few comments that need addressing before we can merge and publish this 😄

Feliz.Recharts/Legend.fs
sydsutton: Here's an example in the Recharts docs. It's a function that can change how a particular legend is formatted. 
image
image
image

Zaid-Ajaj: According to this example, the type of the input function should be `string -> 'a -> ReactElement`, no?

sydsutton: The int is the index, according to the docs. 

Screenshot_20230102_163018

Zaid-Ajaj: if that is the full type, then the lambda should be `string -> 'a -> int -> ReactElement`

sydsutton: Ok got it, fixed!

Zaid-Ajaj (approved):

Zaid-Ajaj:
Thanks a lot @sydsutton for the contribution 🙏! I've merged and published Feliz.Recharts v4.1.0 (also updated Feliz dependency to v2.3.0) 

Just in case you still using Fable v3 + Feliz v1.68, I've published another package v3.16.0 that is compatible with that combination 😄

sydsutton:
Great, thanks so much @Zaid-Ajaj ! I'm pretty new to F#, and development in general, but your continued support of F# is what makes it a great language. I'm just trying to pull my weight. 😉 Thanks again!

Issue #535, opened 2022-12-25
Fable v3 and v4 Compatibility Updates for Feliz v2.0 🚀 


Zaid-Ajaj:
Many changes have been happening across the @fable-compiler and JS ecosystem: 
- Introduction of Fable v4 (new AST to be used by CompilerPlugins)
- The splitting of the Fable.React library into Fable.React.Types + Fable.ReactDom.Types
- The update from React v17 to v18 and its subtle differences with respect to `useEffect`
- Fable.Elmish v4 going stable
- The arrival of .NET7 SDK

Here is a short summary of what works currently in Fable v3 and what needs to happen when updating to Feliz v2 🚀 

### Fable v3 + Feliz v1.68
One might ask: for users of Feliz and its library of components, what still works and what is being worked on?

People using Fable v3.x can keep using the main Feliz library _**up until v1.68**_ and its dependants:
 - [x] Feliz.Recharts v3.15.0
 - [x] Feliz.PigeonMaps v2.7.0
 - [x] Feliz.Popover v2.4.0
 - [x] Feliz.RoughViz v1.6.0
 - [x] Feliz.UseDeferred v1.5.0
 - [x] Feliz.SelectSearch v1.8.0
 - [x] Feliz.Markdown v1.4.0
 - [x] Feliz.Svelte 
 - [x] Feliz.SvelteComponent v0.7.0
 - [x] Feliz.Kawaii v1.3.0 
 - [x] Feliz.UseMediaQuery v1.5.0
 - [x] Feliz.Delay v0.4.0

It is important to note that Feliz v1.68 has these NPM dependencies:
 - [x] `17.0 >= react > 18.0`
 - [x] `17.0 >= react-dom > 18.0`

> ⚠️ Upgrading `react` or `react-dom` to v18+ might break your app. Feliz v1.68 depends on the specific `react` and `react-dom` versions above
> ⚠️ Using Fable v3 with .NET 7 doesn't work! You will need `global.json` file in your project that specifies the SDK version to use `6.0.400`
> ⚠️ Feliz v1.68 _**might**_ work with Fable v4 if you update Feliz.CompilerPlugins to v2

### Feliz.UseElmish v2.x improvements

Going from Feliz.Elmish v1.6.0 (which relied on Fable.Elmish v3 and Feliz) now have been published as **Feliz.UseElmish v2.1.0**:
 - No longer relies on Feliz (pure hook)
 - Should work both with React v17 _and_ React v18
[...]

PR #548, opened 2023-02-06
Move UseListener to separate project

panmona:
I propose that we adopt https://github.com/Shmew/Feliz.UseListener into this project as this is already used in the Feliz docs.and to make it possible to use it with Fable 4. Currently there's a version restriction on the package:
image

I don't think that we'll be able to update the library on Shmew's account, I made a PR nonetheless. https://github.com/Shmew/Feliz.UseListener/pull/4

This would also need some additional build changes, so that the publishing works correctly. Not sure if this would already be done with just adding the new project name [here](https://github.com/Zaid-Ajaj/Feliz/blob/master/build/Program.fs) or if there's more to it.


MangelMaxime:
Hello @panmona,

If your goal is to move ownership of Feliz.UseListener, it is better to ask the original maintainer first about it.

This is because he is the one who should give his consent for that and gives access to the package on NuGet. Indeed, you can't publish to a package that you don't own on NuGet.

panmona:
What I would propose is to publish it under a different name just like with Feliz.Elmish/Feliz.UseElmish, as I'm aware that it isn't possible to publish under the same name. The original author didn't have any sign of activity in the last 2 years.

MangelMaxime:
@panmona I am not sure what yo speaking about Feliz.Elmish doesn't exist.

If you are speaking about Fable.Elmish, then there are 2 packages because they solve different problem.

Let's wait to see if @Shmew is going to answer because you PR is only 1 day old. Or try to contact it via other channel like slack, discord, etc.

We can then decide what to do depending on the situation.

panmona:
@MangelMaxime Sounds good.
I was referencing to this section in #535, maybe the text is wrong?
image

MangelMaxime:
Not what they are speaking about because there are no `Feliz.Elmish` that I can see listed on [Nuget.org](https://www.nuget.org/packages?q=Feliz.Elmish).

But that's not really important :)

Zaid-Ajaj:
@panmona I think this is a good idea. UseListener is a nice library and I was using it in Feliz.Router, it was a bummer that i couldn't use it for v4 of Feliz.Router and had to roll out my own helpers. 

I will have to look at the actual implementation code later. As for the name, something like `Feliz.Listeners` could do, don't really have a strong opinion on the matter

Zaid-Ajaj:
Merged and published under a new name `Feliz.Listeners` with the same implementation :rocket:

panmona:
Thank you!

Issue #544, opened 2023-01-27
ReactComponent attribute does not work without calling `ofFunction` where used


sonicbhoc:
So I am having a strange issue with Feliz 1.68 and Fable 3.7.1.

I'm trying to author React components. When I just add the `[<ReactComponent>]` attribute to an F# function with a capital name and a list of properties (which is usually empty), I'm expecting it to generate a React Function Component. But the generated component's logic is essentially comprised of constructing arrays like what's in the F# code. Which is fine, except my React components don't work.

In every example I've seen, all I am supposed to need is the ReactComponent Attribute and supposedly I'd get a function component, and then I can just call that component like a function. But when I do that, I get this error:

`Invalid hook call. Hooks can only be called inside of the body of a function component.`

And the generated code looks like this:

```
export function Router() {
    let elements;
    const patternInput = Feliz_React__React_useState_Static_1505(RouterModule_urlSegments(window.location.hash, 1));
    const updateUrl = patternInput[1];
    const currentUrl = patternInput[0];
    return RouterModule_router(createObj(ofArray([["onUrlChanged", updateUrl], (elements = toList(delay(() => {
        const matchValue = Url_Parse_1334CEF1(currentUrl);
        switch (matchValue.tag) {
            case 1: {
                return singleton(createElement("h1", {
                    children: ["Logging In..."],
                }));
            }
            case 0: {
                return singleton(Index());
            }
            case 2: {
                return singleton(Estimator());
            }
            case 3: {
                return singleton(TimeTracker());
            }
            default: {
                return singleton(createElement("h1", {
                    children: ["Error 404: Not Found"],
                }));
            }
        }
    })), ["application", react.createElement(react.Fragment, {}, ...elements)])])));
}

export function PageContainer() {
    let elems_5, elms_1, elms, props_2, elems_2, elms_2, props_6, elems_3;
    const props_9 = ofArray([["className", "is-fullheight"], (elems_5 = [(elms_1 = singleton_1((elms = singleton_1(Navbar()), createElement("div", {
        className: "container",
        children: Interop_reactApi.Children.toArray(Array.from(elms)),
    }))), createElement("div", {
        className: "hero-head",
        children: Interop_reactApi.Children.toArray(Array.from(elms_1)),
[...]

PR #554, opened 2023-02-19
Restrict Feliz.CompilerPlugins version required by Feliz 1

MangelMaxime:
Hello @Zaid-Ajaj,

**Important:** This PR is not to be merged into master but should instead go into a feliz1 branch. In theory, I created this branch on the latest commit related to v1.

This branch is related to #544. The goal of this branch is to push a new version of Feliz 1 which restrict Feliz.CompilerPlugin to version 1 only.

image

I had to use a custom MSBuild task because MSBuild doesn't support version restriction on `ProjectReference` out of the box.

https://github.com/NuGet/Home/issues/5556#issuecomment-1179526189

### DIFF
--- a/Feliz/Feliz.fsproj
+++ b/Feliz/Feliz.fsproj
@@ -38,7 +38,21 @@
     <Compile Include="ReactDOM.fs" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\Feliz.CompilerPlugins\Feliz.CompilerPlugins.fsproj" />
+    <ProjectReference Include="..\Feliz.CompilerPlugins\Feliz.CompilerPlugins.fsproj" PackageVersion="[1.1.0, 2.0.0)"/>
   </ItemGroup>
+    <Target Name="UseExplicitPackageVersions" BeforeTargets="GenerateNuspec">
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Feliz v2 has been in use for a while now, I don't think I want to maintain a separate Feliz v1.x branch just for this 🤔 I would rather if users updated to Feliz v2 if they encountered issues with v1

MangelMaxime:
@Zaid-Ajaj The problem, is that if they don't use the latest version of Fable 3 then they will only see the problem at the runtime only with some weird errors / behaviours. There was a bug in previous version of Fable 3 which allows to "skip" invalid plugins and keep the compilation going as normal when it should have failed.

We are speaking about a one time release, which can avoid them having to debug strange bugs.

Freymaurer:
I will close this PR for now. I think going forward from here we will keep branches for the latest release of each major version. 

If you think otherwise @MangelMaxime feel free to reopen, then we will do the work required.

MangelMaxime:
At the time, it would have been nice to have merged but now  that 1-2 years has pass, I think we can avoid doing the work.

PR #557, opened 2023-02-23
Update Fable dotnet tool version

panmona:
This fixes the error with dotnet 7 and also the error with the hooks I reported here: https://github.com/Zaid-Ajaj/Feliz/issues/535#issuecomment-1441418947


Zaid-Ajaj:
That's weird I thought I totally did this yesterday 😓 thanks for spotting it ❤️ merged and published as of Feliz.Template v4.2

PR #558, opened 2023-02-23
Remove `Experimental` attribute from `fitContent` style prop binding

ArtemyB:
Seems like the corresponding `fit-content` CSS property value is already stable: https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content#browser_compatibility


Zaid-Ajaj:
Thanks @ArtemyB for the contribution 🙏 merged and published as of Feliz v2.4.1 🚀

PR #560, opened 2023-02-23
Template Router not working

lukaszkrzywizna:
If I understand the concept of `Feliz.Router` correctly, it should be used as a root of an `App`. Currently, changing the URL gives no result.

### DIFF
--- a/Feliz.Template/Content/src/Main.fs
+++ b/Feliz.Template/Content/src/Main.fs
@@ -5,4 +5,4 @@ open App
 open Browser.Dom
 
 let root = ReactDOM.createRoot(document.getElementById "feliz-app")
-root.render(Components.Counter())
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @lukaszkrzywizna Feliz.Template is just that, a template where config is setup and the rest is up to you. I start with `Counter()` component because everyone knows and if they want to switch to another component, it is up to them.

> Currently, changing the URL gives no result.

That is expected, the root component doesn't implement routing.

lukaszkrzywizna:
Frankly, it's a little bit surprising. The template contains a few components, and the last is the sum of all previous ones enhanced with the routing feature.  I would expect that all features are accessible at the beginning. But I get your point of view of course. 👌

Issue #559, opened 2023-02-23
Feliz.Template v4.x - The namespace or module 'Fable' is not defined


natalie-o-perret:
Hi there, came across a bit of an issue when following the quick-start section:
https://github.com/Zaid-Ajaj/Feliz#quick-start
It seems that, I can't start the template project as one might expect out-of-the-box:

```
#  % natalie-perret @ nppc in ~\Desktop [12:35:56]
$ dotnet new -i Feliz.Template
Warning: use of 'dotnet new --install' is deprecated. Use 'dotnet new install' instead.
For more information, run:
   dotnet new install -h

The following template packages will be installed:
   Feliz.Template

Feliz.Template (version 4.1.0) is already installed, it will be replaced with latest version.
Feliz.Template::4.1.0 was successfully uninstalled.
Success: Feliz.Template::4.1.0 installed the following templates:
Template Name   Short Name  Language  Tags
--------------  ----------  --------  ------------------
Feliz template  feliz       F#        Fable/Elmish/React


#  % natalie-perret @ nppc in ~\Desktop [12:36:01]
$ dotnet new feliz -n MyProject
The template "Feliz template" was created successfully.


#  % natalie-perret @ nppc in ~\Desktop [12:36:07]
$ cd MyProject

#  % natalie-perret @ nppc in ~\Desktop\MyProject [12:36:08]
$ npm install
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@vitejs/plugin-react@3.1.0',
npm WARN EBADENGINE   required: { node: '^14.18.0 || >=16.0.0' },
npm WARN EBADENGINE   current: { node: 'v15.14.0', npm: '7.7.6' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'rollup@3.17.2',
npm WARN EBADENGINE   required: { node: '>=14.18.0', npm: '>=8.0.0' },
npm WARN EBADENGINE   current: { node: 'v15.14.0', npm: '7.7.6' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'vite@4.1.4',
npm WARN EBADENGINE   required: { node: '^14.18.0 || >=16.0.0' },
npm WARN EBADENGINE   current: { node: 'v15.14.0', npm: '7.7.6' }
npm WARN EBADENGINE }

[...]

PR #562, opened 2023-02-26
Enforce Node.js version in Feliz.Template

MangelMaxime:
Related to discussion in #559


Zaid-Ajaj:
Thanks a lot @MangelMaxime merged and published as of Feliz.Template v4.3.0 🚀

PR #563, opened 2023-03-03
Set displayName for ReactMemoComponent

### LINKED ISSUE #561: ReactMemoComponent attribute and displayName
Currently, using `ReactComponent` causes generating component with `displayName` set to component's name (JS function name):
image
However, this is not the case for `ReactMemoComponent`
image

Regarding https://julesblom.com/writing/component-displayname Feliz would have to:
- Use named function declaration as a memo's parameter 
or
- Set `displayName` manually.

Is this something possible to implement?

lucassouza1:
Fixes #561 

image


Zaid-Ajaj:
Thanks a lot @lucassouza1 for the fix! I've merged and published this as of Feliz.CompilerPlugins v2.1.0 🚀 which is now included as of Feliz v2.5.0

PR #568, opened 2023-03-20
let the doc comment for imported render correctly

dawedawe:
This let's doc comment about the attribute render correctly.
image


Zaid-Ajaj:
Thanks @dawedawe merged and published as part of Feliz v2.6.0 🚀

PR #580, opened 2023-09-04
ReactListComponent POC

lukaszkrzywizna:
POC for #581


Freymaurer:
Hello! My guess is this PR was intented as another solution for static vs dynamic children? Or as an alternative to React.fragement?

As we will go forward using jsx syntax, this should become obsolete. I will close this PR for now.

If you think there is still merit in this PR, feel free to contact me and we can decide on how to move forward.

PR #583, opened 2023-09-24
Addition of Scatter chart

mrtz-j:
This PR adds the ReCharts.ScatterChart and an entry into the live-samples based on a translation of the original documentation. 

In addition I also added some of the missing props for `XAxis`, `YAxis` and `scatter` from the Recharts API page.


Zaid-Ajaj:
Thank you @mrtz-j merged and published as of v4.2.0 🎉

Issue #584, opened 2023-09-30
CSS overflow-anchor property


pw242:
I can't find the CSS overflow-anchor property anywhere.
https://www.w3schools.com/cssref/css_pr_overflow-anchor.php

MangelMaxime:
Hello @pw242,

It indeed seems to be missing.

Would you like to try send a PR?

I believe you should be able to use `textOverflow` as a reference: https://github.com/Zaid-Ajaj/Feliz/blob/1b7298ebd6ec3a7d5eb1eb7b3aa71da7ae2520f9/Feliz/Styles.fs#L3640-L3650

PR #585, opened 2023-09-30
Implement overflow-anchor style property.

pw242:
#584 First attempt at pull request as suggested.


Zaid-Ajaj:
Thank you @pw242 merged and published as of Feliz v2.7.0 🎉

PR #587, opened 2023-10-20
Making UseElmish's dispatch function stable.

### LINKED ISSUE #588: Making UseElmish's dispatch function stable
Hi @Zaid-Ajaj! 
During running my application I've encountered a problem with `React.UseElmish`. When a component using `React.memo` with custom equals function (omitting functions), the dispatch provided from `UseElmish` simply doesn't work. The issue is caused by replacing the `initialDispatch` function with a real one, provided by `Elmish's` program. We can simply fix this by making dispatch stable - hook would always reuse `initialDispatch` but would check its logic when the Program's dispatch is provided. The code below provided the discussed scenario:

```fsharp
type Model = {
  Counter: int
}

type Msg =
  | Increase

let init () = { Counter = 0 }, Cmd.none
let update msg model =
  match msg with
  | Increase -> { model with Counter = model.Counter + 1 }, Cmd.none

let MButton =
  React.memo((fun (props: {| counter: int; dispatch: Msg -> unit |}) ->
    Html.button [
      prop.text $"Counter: {props.counter}"
      prop.onClick (fun _ -> props.dispatch Increase)
    ]),
    areEqual = Fable.React.Helpers.equalsButFunctions // this will force to use initial dispatch function
  )

let App() =
  let state, dispatch = React.useElmish(init, update, [||])
  MButton {|
    counter = state.Counter
    dispatch = dispatch
  |}

```

Note: I did not provided any test for it. I have a problem with running `Feliz` project and tests. Do you have any manual how to run it?

lukaszkrzywizna:
Resolves #588


MangelMaxime:
Hello @lukaszkrzywizna,

Thanks for the PR

Did you test if your changes support HMR? I am asking because this is something important to keep supporting.

lukaszkrzywizna:
I've tested it using the `Vite` + `React Fast Refresh` plugin. I've checked if during updating components:
- the state persist
- `Vite` doesn't notify about `HMR` problems. 

Is that it? Should I check something else?

MangelMaxime:
> * the state persist

If the state persist and the application is not broken then that's good enough in general.

> * `Vite` doesn't notify about `HMR` problems.

I believe in the browser console you have a log. But in general, if you don't have a full refresh of the page it means that HMR happened.

lukaszkrzywizna:
@Zaid-Ajaj Did you have a change to look at this?

Zaid-Ajaj:
@lukaszkrzywizna the changes look reasonable, I haven't had the chance to actually try them on a project though and the problem is I believe we have done this _fix_ already a few times and every time something changes in tooling like vite or fast refresh and we need another fix. 

If you say this works with latest tooling, we can get this merged and publish a package ✅ if we hear it breaks things we can revert. Hard to unit test things like this 😓

Zaid-Ajaj:
Thanks @lukaszkrzywizna for the PR 🙏 ❤️ merged and published as of Feliz.UseElmish v2.5.0 🚀

lukaszkrzywizna:
> If you say this works with latest tooling, we can get this merged and publish a package ✅ if we hear it breaks things we can revert. Hard to unit test things like this 😓

Sorry for the late response. I've been using it in a production for about 3 weeks. No problems so far 🤞 

> Thanks @lukaszkrzywizna for the PR 🙏 ❤️ merged and published as of Feliz.UseElmish v2.5.0 🚀

Thank you for merging it! I'm glad I could add something to this awesome library 😄

Issue #564, opened 2023-03-03
Cluttered imported React hook names


ArtemyB:
Here are the examples of React hook calls currently (`Feliz v2.4.1`, but I remember the case take place all my `Feliz` usage): `useReact_useCallback_1CA17B65`, `useReact_useMemo_10C6A43C` and so on. Firstly, That doesn't look pretty. And in practice it affects React Dev Tools user experience. Secondly, there is actual JS-code being produced in Feliz bundle, e.g. from `fable_modules/Feliz.2.4.1/React.fs.js`:
```js
export function useReact_useMemo_10C6A43C(createFunction, dependencies) {
    const arg_1 = defaultArg(dependencies, []);
    return Interop_reactApi.useMemo(createFunction, arg_1);
}
// and so on for the rest of the bindings
```

As I understand, this would not have happened if the containing [`React` type](https://github.com/Zaid-Ajaj/Feliz/blob/a0fca744c584596728ccf8971a57de2ba9471ef2/Feliz/React.fs#L59) had `Erase` attribute (to be erased by Fable). However it will work only when all of the type members are `inline`. And now I see that alongside the `inline` "pure" imports [`React`](https://github.com/Zaid-Ajaj/Feliz/blob/a0fca744c584596728ccf8971a57de2ba9471ef2/Feliz/React.fs#L59) type contains:
- non-`inline` members that in fact could be made `inline` because of being "pure" imports);
- non-`inline` members and even custom hooks with the actual logic (`createDisposable`, `useEffectOnce`, `useCallbackRef`, `useCancellationToken`) that couldn't be simply marked `inline` while being defined directly in the `React` type as they're now. However I see the solution in moving all these custom hooks to a separate module/non-erased type and then calling them from the `inline` members of `React` type -- so that there should be no critical changes from the F# API perspective.
P.S. Also `Hook` attribute should be removed from all these hook-members when the `React` type is erased.

The solution doesn't seems to require a lot of work, if it is actually correct -- haven't tried anything yet, just made assumptions based on my previous experience. Decided to ask for other opinions and expertise here before doing something 🙂

Zaid-Ajaj:
Hi @ArtemyB I understand the generated JS is not ideal by Feliz, however sometimes it is not possible to inline members. I don't remember the exact reasons why some hooks aren't being inlined. That said, a PR would be really appreciated that includes these improvements 🙏

PR #591, opened 2023-12-24
Refactor React hooks and APIs F# API to improve the transpilation output

ArtemyB:
Solves #564

I haven't changed any publicly oriented F# API, i.e., module, type and function names are preserved. The only change kind of affects the public API is switching `Feliz.ReactInterop` and `Feliz.Internal` access modifiers to public , whereas, I suppose, they're not intended to be a part of the public F# API. But otherwise their functions cannot be referenced from the public inline functions of the `React` type that is an implied public API.

Furthermore, I haven't run any automated tests yet (don't know if they're up-to-date, actually). However, I've tried the new version with one of my projects that uses many of the APIs, and I haven't noticed any problems so far. At the same time, the hooks summary in the React Developer Tools did improved, i.e., there's no more entries with names like `useReact_useMemo_...` and with redundant nested hooks. Instead, there are clear hooks as they were written manually in JS.

The changes description:

- Move `Feliz.DateParsing`, `Feliz.Helpers`, `Feliz.ReactInterop` modules to separate files.
- Convert `Feliz.Internal` internal type into a module and move it to a separate file. Make the module public to allow its definitions being called from the public `React` type's inline members.
- Adjust `useEffect` method definition in `Feliz.ReactApi.IReactApi` interface.
- Change `ReactInterop` binding definitions to the attribute-based to ensure the correct import statements generated by Fable (otherwise, the path to "ReactInterop.js" will be relative to the "Feliz" package root and not to the root of the referencing site).
- Make `Feliz.Helpers.optDispose` function not inline - to reduce the amount of Feliz transpiled code.
- Make all the members of `Feliz.React` type inline, and for those that contain some custom logic (beyond the trivial bindings) move the logic into the separate functions of the `Feliz.Internal` module. This way, the `React` type won't be present in the transpiled code.
- Switch `Feliz.ReactInterop`  module access modifier from `internal` to `public` to allow its definitions being called from the public `React` type's inline members.

### DIFF
--- a/Feliz/DateParsing.fs
+++ b/Feliz/DateParsing.fs
@@ -1,0 +1,52 @@
+﻿module [<RequireQualifiedAccess>] Feliz.DateParsing
+
+open System
+
+let (|Between|_|) (x: int) (y: int) (input: int) =
+    if input >= x && input <= y
+    then Some()
+    else None
+let isLeapYear (year: int) = DateTime.IsLeapYear(year)
+
+let (|Int|_|) (input: string) =
+    try (Some (int input))
+    with _ -> None
+/// <summary>Parses a specific yyyy-MM-dd or yyyy-MM-ddTHH:mm date format that comes out of an input element with type="date"</summary>
+let parse (input: string) : Option<DateTime> =
+    try
+        if String.IsNullOrWhiteSpace input then None
+        else
+        let parts = input.Split('-')
+        let year, month, day, hour, minute =
+            match parts with
+            | [| Int year; Int month |] -> year, month, 1, 0, 0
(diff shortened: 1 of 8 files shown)


Zaid-Ajaj:
Thanks a lot @ArtemyB for the this change, it looks reasonable to me. I'll merge and publish sometime this week 🙏

Zaid-Ajaj:
Actually never mind, since this is a beta build, I've merged and published it as of 2.8.0-beta.0 🚀 please take it for a spin and if there are no problems we can publish a proper v2.8.0 version sometime soon 😄 thanks again for the PR 🙏

ArtemyB:
Great, thanks 👍 Then I'll continue using it on my projects and observe 👀

PR #600, opened 2024-03-26
Refined `useEffectOnce` to solely act as an alias for `React.useEffect(effect, [||])`.

### LINKED ISSUE #595: Clarification and Potential Improvement on PR #480 useEffectOnce ehavior
@Zaid-Ajaj @JordanMarr I've checked the PR #480 and I wonder if that code is correct when **not** using strict mode:

```fsharp
if renderAfterCalled.current
then destroyFunc.current
else None
```
I'm asking because if I an component unmounts the destroy func is not invoked:

```fsharp
// not using strict mode

let dispose = React.createDisposable(fun () -> eprintf "Disposed")
    
// dispose never invoked
React.useEffectOnce(fun () -> dispose)
    
// disposed invoked
React.useEffect((fun () -> dispose), [||])
```

I wonder what's the point of this function? If it's created to handle strick mode's double useEffect run, then we should reflect this in a name or comment. Otherwise we should fix it somehow - my first idea is to add check for a debug:

```fsharp
  static member private useEffectOnce(effect: unit -> System.IDisposable) = 
    let destroyFunc = React.useRef None
    let calledOnce = React.useRef false
    let renderAfterCalled = React.useRef false

    if calledOnce.current then
        renderAfterCalled.current <- true
    
    React.useEffect (fun () -> 
        if calledOnce.current 
        then None
        else
            calledOnce.current <- true
            destroyFunc.current <- effect() |> Some
#if DEBUG
            if renderAfterCalled.current
#else
            if not renderAfterCalled.current
#endif
            then destroyFunc.current
            else None
    , [||])
```
But this does not guarantee that user has strict mode enabled and we can still have situation with not invoked dispose.

What do you think about it?

lukaszkrzywizna:
Resolves #595

### DIFF
--- a/Feliz/Internal.fs
+++ b/Feliz/Internal.fs
@@ -87,7 +87,7 @@ module Internal =
                 calledOnce.current <- true
                 destroyFunc.current <- effect() |> Some
 
-                if renderAfterCalled.current
+                if not renderAfterCalled.current
                 then destroyFunc.current
                 else None
         ) [||]
@@ -108,7 +108,7 @@ module Internal =
                 calledOnce.current <- true
                 destroyFunc.current <- effect()
 
-                if renderAfterCalled.current
+                if not renderAfterCalled.current
                 then destroyFunc.current
                 else None
         ) [||]
--- a/Feliz/Internal.fs
+++ b/Feliz/Internal.fs
@@ -63,55 +63,15 @@ module Internal =
 
     [<Hook>]
     let useEffectOnce(effect: unit -> unit) =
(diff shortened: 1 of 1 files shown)


Freymaurer:
Hello! I somewhat agree with the changes proposed here. Today, while revivifing the test suit, i noticed that the test about useEffect were failing in some cases #655. I added referenced to this PR and the relevant issue to #655 and will resolve this after updating to the jsx output as suggested by Maxime.

Are you willing to rebase the PR when the project restructure is more advanced? Otherwise i will just implement the required changes.

PR #606, opened 2024-05-23
Added wrapping object for every type-declared record used as a prop object with `[<ReactComponentAttribute>]`

### LINKED ISSUE #603: Enhance Handling of F# Record Types as Props in ReactComponent Attribute with Warnings, Blocking, or Automatic Wrapping
### Issue:
When using Feliz with F# and Fable, passing F# record types as props in React components results in unexpected behavior. The record type, which Fable translates into a JavaScript object with a specific prototype, loses its prototype once processed by React.createElement. This leads to the absence of expected methods and properties on the prototype.

_Illustration of simple record transformation:_

image

_Differences in objects with and without the ReactComponent attribute:_

image

### Current Workaround by Users:
Currently, a practical workaround for users is to manually wrap the record into an anonymous object. This method involves encapsulating the record as a single property within an anonymous props object, which preserves the prototype through the React component lifecycle.

_Example of wrapping the record:_

image
image

### Proposed Solution for Feliz:
I propose that Feliz should either:

1. Automatically wrap F# record types in an anonymous object when detected as props, thus preserving the prototype integrity.
2. Implement a warning or error system that alerts developers when a record type is used as props, guiding them towards safer practices or the recommended workaround.

@Zaid-Ajaj @MangelMaxime I'm interested in what you think.

lukaszkrzywizna:
It resolves #603

### DIFF
--- a/Feliz.CompilerPlugins/ReactComponent.fs
+++ b/Feliz.CompilerPlugins/ReactComponent.fs
@@ -102,29 +102,30 @@ type ReactComponentAttribute(?exportDefault: bool, ?import: string, ?from:string
                 // declared record
                 // https://github.com/Zaid-Ajaj/Feliz/issues/603
                 // F# Component { Value = 1 }
-                // JSX <Component props={ { Value={1} } } />
-                // JS createElement(Component, { props = { Value: 1 } })
+                // JSX <Component props={ Value={1} } />
+                // JS createElement(Component, props = { Value: 1 })
                 
                 // anonymous record
-                // F# Component { Value = 1 }
+                // F# Component {| Value = 1 |}
                 // JSX <Component Value={1} />
                 // JS createElement(Component, { Value: 1 })
                 
                 let isDeclaredRecord = AstUtils.isDeclaredRecord compiler info.Args[0].Type
                 
-                if AstUtils.recordHasField "Key" compiler info.Args[0].Type then
+                match AstUtils.tryGetRecordField "key" compiler info.Args[0].Type with
+                | Some keyField when isDeclaredRecord ->
                     // When the key property is upper-case (which is common in record fields)
                     // then we should rewrite it
                     let modifiedRecord =
-                      if isDeclaredRecord then
-                        AstUtils.objExpr [
-                          "key", AstUtils.emitJs "$0.Key" [info.Args[0]];
-                          membArgs[0].Name.Value, info.Args[0]
-                        ]
-                      else
-                        AstUtils.emitJs "(($value) => { $value.key = $value.Key; return $value; })($0)" [info.Args[0]]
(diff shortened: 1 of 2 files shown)


Feliz.CompilerPlugins/ReactComponent.fs
```diff
@@ -74,16 +98,39 @@ type ReactComponentAttribute(?exportDefault: bool, ?import: string, ?from:string
...
             if List.length membArgs = info.Args.Length && info.Args.Length = 1 && AstUtils.isRecord compiler info.Args[0].Type then
+                
+                // declared record
+                // https://github.com/Zaid-Ajaj/Feliz/issues/603
+                // F# Component { Value = 1 }
+                // JSX <Component props={ { Value={1} } } />
+                // JS createElement(Component, { props = { Value: 1 } })
+                
+                // anonymous record
                 // F# Component { Value = 1 }
```
MangelMaxime, line 109: For clarity we could perhaps use the anonymous syntax ?

```fs
// F# Component {| Value = 1 |}
```

Zaid-Ajaj:
Hi @lukaszkrzywizna I know it's been a while since we implemented the record wrapping in its current state. I can't seem to remember why exactly it was needed 🤔 maybe for cases where React was re-rendering things because the record got a new reference but the data hasn't changed.  

Personally I feel like we shouldn't change anything here. It's hard to maintain F#-ness of things when going into React world. Keep things simple and when you need F#-ness maintained, use anonymous record to wrap things. Also before adding something like this, we would need to revive the unit test suite to ensure we are not breaking anyone with these changes 😅

lukaszkrzywizna:
Thank you @Zaid-Ajaj for the response!

> React was re-rendering things because the record got a new reference but the data hasn't changed.

Hmm... I don't recognize any mechanism similar to this. Maybe React.memo? But according to [documentation](https://react.dev/reference/react/memo) it compares props one-by-one, not the whole props object.

> Personally I feel like we shouldn't change anything here.

I would agree, but tracking down the issue in my production app cost me many hours. The record passed as an argument was used in a dispatch message where the handler retrieved it and checked if the map contained it. The entire record prototype was erased, so it couldn't work properly, causing the problem to manifest far away from the actual cause in the app. Any information from the compiler would be a huge help in the future, but I understand that changing the behavior of record construction could be significant. That's why I propose at least showing a warning.

> we would need to revive the unit test suite to ensure we are not breaking anyone with these changes

Do you mean the 'Tests' project? I can try to revive this. Do you know what should be done?

Zaid-Ajaj:
> Do you mean the 'Tests' project? I can try to revive this. Do you know what should be done?

Yeah that's the one. mostly going for github actions instead of appveyor, updating deps etc.

PR #609, opened 2024-06-04
Add more CssUnit options

Linschlager:
Add min, max, minmax, minContent, fitContent, maxContent and custom length options.

This should help making tricky css grid code a bit nicer.


Zaid-Ajaj:
Included in Feliz v2.8 🚀

PR #614, opened 2024-06-04
Add webkitLineClamp style

Linschlager:
> [!WARNING]
> This is non-standard API but currently supported in all major browsers. It's up for discussion if this should be supported in Feliz

Adds styles:
- display: -webkit-box and -webkit-inline-box
- -webkit-box-orient
- -webkit-line-clamp

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -2657,7 +2657,15 @@ type style =
             unbox<string> xOffset + " " +
             unbox<string> yOffset + " " +
             unbox<string> zOffset
-        ) 
+        )
+    
+    /// Limits the contents of a block to the specified number of lines.
+    /// 
+    /// It only works in combination with the `display` property set to `webkitBox` or `webkitInlineBox`
+    /// and the `webkitBoxOrient` property set to vertical.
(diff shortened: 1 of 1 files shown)


Freymaurer:
> This is non-standard API but currently supported in all major browsers. It's up for discussion if this should be supported in Feliz.

I too am conflicted here. It seems to rely on some deprecated properties and i am unsure about the browser support. Might be me being unable to comprehend what bottom `none` exactly means.

> Note: For legacy support, the vendor-prefixed -webkit-line-clamp property only works in combination with the [display](https://developer.mozilla.org/en-US/docs/Web/CSS/display) property set to -webkit-box or -webkit-inline-box and the [-webkit-box-orient](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient) property set to vertical. Despite these prefixed properties being deprecated, the co-dependency of these three properties is a fully specified behavior and will continue to be supported.

https://developer.mozilla.org/en-US/docs/Web/CSS/line-clamp#browser_compatibility

Will close this PR for now. If you disagree feel free to reopen then we can discuss on how to proceed

PR #615, opened 2024-06-04
Add outline shorthand property

Linschlager:


Zaid-Ajaj:
Included in Feliz v2.8 🚀

PR #622, opened 2024-09-15
Fix Markdown link (a) component

tboby:
This confused me for a while as I thought I was using `components` wrong!

```fsharp
 Markdown.markdown [
      markdown.children rel.Description
      markdown.components [
          unbox<IComponent> (
              Interop.mkAttr "a" (fun (x: ILinkProperties) ->
                  Html.a [
                      prop.children x.children
                      prop.className "link"
                      prop.href x.href
                  ])
          )
      ]
  ]
```
works, which I assume is what `linkReference` is supposed to be overriding.

The table at https://github.com/remarkjs/react-markdown/blob/main/changelog.md#600---2021-04-15 implies that trow/tcell may be wrong too, but I didn't check that.


Freymaurer:
Hello @tboby ! If you are still using Feliz.Markdown i would really appreciate your input on this 🙂 

I update project structure on [feliz_v3.0](https://github.com/fable-hub/Feliz/tree/feliz_v3.0), can you create a pr with your fix against that branch AND possibly create a first test for the Feliz.Markdown project? 

I will gladly stay in contact with you if you have any questions.

Issue #626, opened 2024-10-21
Docs conversion


Freymaurer:
Hey! I use Feliz since for a lot of projects and wanted to contribute to the documentation, but nothing was working for me 😅 

I checked the documentation project and saw that it was very complex with ~1400 lines of code for routing alone.

Combined with a comment i found, i came here with a suggestion.

> Hi there @Darkle, apologies for the delay. I've merged your PR but the docs application is broken and needs an update. Probably an overhaul along with the changes to the CI. I will be fixing it as soon as time permits 🙏

_Originally posted by @Zaid-Ajaj in https://github.com/Zaid-Ajaj/Feliz/issues/618#issuecomment-2184162297_

## Goal

My goal was using a well developed documentation framework, while displaying f# code and the resulting react components.

## Implementation

I forked this repo and set up a [astro](https://astro.build) [starlight](https://starlight.astro.build/de/) documentation in the `public` folder.

> [!note]
> Astro can render any frontend framework components (react, svelte, vue,...)

1. Create a f# project in `public/src/components/fsharp`.
2. Update `npm run dev` to include fable transpilation `dotnet fable watch src/components/fsharp -e jsx --run astro dev`
3. Add dependency to Feliz nuget package (the Feliz.fsproj did not work for me. There were transpilation errors vor Feliz.CompilerPlugins and i wanted to get some feedback before investigating any deeper issues)
4. Create POC react components in f# in `public/src/components/fsharp` 
5. Add Feliz core documentation to `public/src/docs/Feliz`
6. Showcase code and component import for example in `public/src/docs/Feliz/ReactApiSupport.mdx`

```mdx
import rawCounter from '/src/components/fsharp/Counter.fs?raw';
import {Counter} from '/src/components/fsharp/Counter.jsx';

<Code code={rawCounter} lang='fsharp'/>
<Counter client:only/>
```

![image](https://github.com/user-attachments/assets/9e844ab4-c98d-4111-baa0-355dd3cc9d11)

## Advantages

- it just works™️ ; Astro works well and is highly customizable. Far less code to maintain. No routing logic, etc.
- Display actual code used for component next to the component. Feature changes/obsolete stuff gets caught easily.

## Disadvantages

- No full f#?

## Conclusion

The current documentation setup is very complicated and can create a entry barrier for new contributers. My suggestion offers an easy alternative to a full f# documentation project, reusing existing features from astro while focusing on f# components.

PR #627, opened 2024-10-21
Docs conversion

Freymaurer:
See #626 for more information


Zaid-Ajaj:
Hi there @Freymaurer first of all, this is looking freaking awesome! thanks a lot 🙏 ❤️ it's true that routing was very complicated because we were mixing markdown with Feliz components for better docs but essentially that is what `.mdx` is for 😄 I haven't tried the PR yet locally but I will have some time soon to check it out! Thanks again 🚀

PR #631, opened 2024-10-22
Fix scrollbarColor

DashieTM:
scrollbarColor with a single string is not supported: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
Instead it just allows the "auto" keyword.

### DIFF
--- a/Feliz/Styles.fs
+++ b/Feliz/Styles.fs
@@ -3135,15 +3135,6 @@ type style =
     /// The thumb refers to the moving part of the scrollbar, which
     /// usually floats on top of the track.
     /// https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
-    static member inline scrollbarColor (color: string) =
-        Interop.mkStyle "scrollbarColor" color
-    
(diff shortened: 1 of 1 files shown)


DashieTM:
Oh oops, sorry the code doesn't show the overloaded function.
There were two versions of this function, one with a single string and one with two. 
You can ofc just pass both in one string, but this feels error prone, where as the second would allow you to properly pass two values and make it clear to the user as well.

Perhaps I can instead just add some extra documentation.

Zaid-Ajaj:
> Perhaps I can instead just add some extra documentation.

Yes, please 🙏 if you really want to make less error-prone, then another overload that takes two string colors makes it more explicit

DashieTM:
This is on master.
image
What I meant is that the function with two parameters already existed alongside the one with one parameter.
E.g. no more changes would be necessary.

Zaid-Ajaj:
Merged and published as of Feliz v2.9 🚀 along with your other PRs. Thanks a lot 🙏

PR #639, opened 2024-12-26
Add overloads for props that require string values for FS.FluentUI

sydsutton:
In FS.FluentUI, Microsoft's `Theme tokens` have pre-determined values for properties like fontSize, lineHeight, borderRadius, strokeWidth, zIndex, boxShadow, etc., but they are all strings. Since a lot of these Feliz styles don't have string overloads for these properties, they are unusable. For example, when using `Theme.tokens.borderRaduisSmall`, the library adds `border-radius: var(--borderRadiusSmall);` to the CSS styling via a className, which is then translated into `border-radius: 2px;`. 

Ideal usage example: `card.style [ style.boxShadow Theme.tokens.shadow64 ] ]` (Theme.tokens.shadow64 being a string)
If there's a more library-specific way that I can accomplish this, please let me know.

Here's part of the type for FluentUI's Token object for further reference: 
```fsharp

type Tokens = {
    borderRadiusNone: string
    borderRadiusSmall: string
    borderRadiusMedium: string
    borderRadiusLarge: string
    borderRadiusXLarge: string
    borderRadiusCircular: string
    fontSizeBase100: string
    fontSizeBase200: string
    fontSizeBase300: string
    fontSizeBase400: string
    fontSizeBase500: string
    fontSizeBase600: string
    fontSizeHero700: string
    fontSizeHero800: string
    fontSizeHero900: string
.... 
}
```

### DIFF
--- a/Feliz/Properties.fs
+++ b/Feliz/Properties.fs
@@ -651,6 +651,9 @@ type prop =
     /// SVG attribute to define the size of the font from baseline to baseline when multiple
     /// lines of text are set solid in a multiline layout environment.
     static member inline fontSize (value: int) = Interop.mkAttr "fontSize" value
+    /// SVG attribute to define the size of the font from baseline to baseline when multiple
+    /// lines of text are set solid in a multiline layout environment.
+    static member inline fontSize (value: string) = Interop.mkAttr "fontSize" value
 
     /// A space-separated list of other elements’ ids, indicating that those elements contributed input
     /// values to (or otherwise affected) the calculation.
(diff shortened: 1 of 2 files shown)


shayanhabibi:
Will be interested to hear thoughts from Zaid, [but as discussed with Syd](https://github.com/sydsutton/FS.FluentUI/pull/18#issuecomment-2563194838), I think this is something that should be provided extension by consumer libraries rather than detracting from strong typing of props

sydsutton:
> I think this is something that should be provided extension by consumer libraries rather than detracting from strong typing of props

Yeah, I agree. I'll add them in an FS.FluentUI-specific manner. Please disregard this PR. Thanks!

Issue #21, opened 2019-08-18
Why React.Children.toArray in prop.children?


cmeeren:
Is the use of `React.Children.toArray` necessary in `prop.children`?

https://github.com/Zaid-Ajaj/Feliz/blob/51ec0c05679a0c3e11f72506b8484b23abd84514/src/Properties.fs#L176

It seems to work fine without. The only difference is that `toArray` seems to add a key to each child element, so you don't get the warning that arrays are supposed to have keys, which may be a bad thing if you're supposed to get that warning.

Looking at the docs for [Children.toArray](https://reactjs.org/docs/react-api.html#reactchildrentoarray), its use-case seem to be flattening more complex structures, and adding keys is a side-effect of this. So it doesn't seem to be useful for this context.

I'm all very new to this low-level stuff though and may be wrong.

zanaptak:
Without this, we get the key warning on every multi-child element in the entire app (see #6), which is arguably worse than no warnings as it becomes overwhelming noise to the dev experience at that point, and it would be unnecessary boilerplate to have to put manual keys on everything.

However it could be argued that the warning is useful in many cases. If the user is adding/removing elements conditionally, or reordering them, then it would be nice to have the warning since those are the situations where you want a deliberate key to assist the diffing algorithm. But how would a library know when to provide the warning and when not?

With the old DSL there was the recommendation to always use `ofList` instead of `yield!`, and `ofList` always gives the warning to use keys. However, now there is some hurdle in using `ofList` (need to mix DSLs), and if you do use it, it no longer produces the warning. Users might be inclined to return to using `yield!`. So in some sense maybe it is an overall regression to hide the warning even despite all the noise. Maybe an `ofList` equivalent needs to be brought into this library.

Or maybe there could be additional helpers to nudge users into correct `key` usage. For example, what if the warning was brought back to `prop.children` by default, but also provide a `prop.staticChildren` (or some better name) that auto-generates the keys to remove the warning. Now, the user can explicitly choose how to implement the keys. If they have static list of elements that won't need diff optimization. they can make a deliberate choice to auto-generate keys with `staticChildren`. But, otherwise the "normal" `prop.children` will remind them to use keys for their dynamic children, and maybe even `yield!` is safe to use again because now they are adding keys where necessary.

cmeeren:
[...]

PR #653, opened 2025-03-19
Proposal for representing static children and selectively avoid React warnings

Canna71:
This PR aims at solving the issue of not currently having a way to distinguish a static list of children from a dynamic one (e.g. a List.map).
Any solution that suppresses React warnings for lists without key is IMHO bad and it would be better to provide a way to express a static list of children; this should not be the default and using it will be a conscious choice, so we will never hide the warnings to the developers so they will realize when they missed to pass the keys.
More details and the logic behind it are in the discussion of the Issue #21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,5 +209,6 @@ dist
 # JetBrains Rider
 .idea
 *.fs.js
+*.fs.js.map
 0.bundle.js
 .yarn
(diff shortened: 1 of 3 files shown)


Freymaurer:
Hello @Canna71 ! We are currently restructuring Feliz for a v3.0.0 release. This release will shift Feliz to output JSX syntax which will circumvent the issue of static/dynamic children and handle them correctly!

I will add a module/docs guide for people who still want to use `createElement` syntax and will factor in your code here.

PR #656, opened 2025-04-30
Project restructure

Freymaurer:
- Move testsuit to vitest, react-testing-library and jsdom 
  - remove need for webpack, copy pasting fable output to different folders, 
- Use bun as dependency manager.
  - Allows manual adding of dependencies for each project.
  - Can run all test projects with a single command from root.
- Seperate test projects into areas of concern. Move into `tests/` :truck:
- Move all feliz projects into `src/` :truck:
- Add Feliz.Vitest.React project with minimal bindings for vitest
- Update some Fable.ReactTestingLibrary bindings to return the correct Promise output, after moving the bindings into their own fsproject :truck:
- remove paket package manager 🔥 
- remove travis/appveyor in favor of github actions (gh actions still must be added)
- fully remove fake build project 🔥 

The goal of this PR was to reduce the overall repo complexity. Due to the nature of this there are a lot of files moved which leads to almost all files being changed.


package.json
```diff
@@ -1,46 +1,30 @@
...
+  "private": true,
+  "scripts": {
+    "publish-docs": "npm run build && node publish.js",
+    "start": "dotnet tool restore && dotnet fable watch docs --noCache --exclude Feliz.CompilerPlugins --run webpack-dev-server",
+    "build": "dotnet tool restore && dotnet fable docs --noCache --exclude Feliz.CompilerPlugins --run webpack --mode production",
+    "test": "bun -F ./tests/* test"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Zaid-Ajaj/Feliz.git"
```
slyshykO, line 11: Think this URL should be changed

Freymaurer: Good catch! Will update!

Directory.Build.props
```diff
@@ -1,14 +1,15 @@
...
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PackageIcon>fable_logo.png</PackageIcon>
+    <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
+    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)\fable_logo.png" Pack="true" PackagePath="\" />
+    <None Include="$(MSBuildThisFileDirectory)\README.md" Pack="true" PackagePath="\"/>
+  </ItemGroup>
+</Project>
```
MangelMaxime, line 15: Nit picking, but I would keep using 4 spaces indentation as I find it easier to read

bun.lockb
MangelMaxime, line 1: I think I would prefer to keep using Node.js instead of Bun because most people will use Node.js and I am not sure if they are bugs or issues that could go un-noticed because of Bun move.

It also make one less tool to install on contributors machines as most will probably have Node.js already

paket.dependencies
MangelMaxime, line 1: Because there are a lot of packages I would probably keep using `paket` to make updating dependencies easier.

If we were to not use `paket`, it would probably be a good idea to use `Central Package Management` from NuGet in order to align the dependencies version easily.

With the current setup, we need to manually update for each packages the required version of FSharp.Core, Fable.Core, etc. which could easily be made out of sync.

I played a lot with NuGet CPM, but I am still unsure what to think about it. I like the idea of having one less tool in the invocation chain, but not being able to do things like `dotnet paket update` or having to find workaround to not have `tests` dependencies impact `src` dependencies is annoying.

All that to say, I probably prefer `paket`

tests/vitest-setup.ts
```diff
@@ -0,0 +1,7 @@
+import '@testing-library/jest-dom/vitest';
+import {cleanup} from '@testing-library/react'
+import {afterEach} from 'vitest'
+
+console.log("vitest-setup.ts loaded")
```
MangelMaxime, line 5: Should we keep the log?

tests/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from 'vitest/config'
+import react from '@vitejs/plugin-react';
+
+console.log('loading vitest config');
```
MangelMaxime, line 4: Should we keep the log?

tests/vitest.config.ts
MangelMaxime, line 1: Would be nice to configure Github CI to replace Travis/AppVeyor.

Should we do it in this PR or another one?

MangelMaxime (changes requested):
@Freymaurer I left some comments, I mark the PR as red mainly to ensure we have discussion about the different topics.

paket.dependencies
Freymaurer: I removed paket, because it was only used for 3 project. The build project, the single test project that existed and Feliz itself. All other projects referenced dependencies directly in the respective .fsproj file.

Personally i really dislike paket, for the same reason as you mentioned with bun. It's just another tool and can lead to unexpected issues.

That said, my first idea was moving to CPM. I had everything setup but could not get it to run. After trying around for 3 hours i scrapped it. But this issue was less relevant than i thought, as most packages in Feliz did not have references to the local Feliz.fsproj but instead to the Feliz nuget package (and to different versions at that).

Feel free to give me some feedback, as i also was unsure if we should change these Feliz dependencies to the local package.

tests/vitest-setup.ts
Freymaurer: Personally i like having that log to see if it **really** loads the config files 😅 But i can remove them if we are done moving parts around

tests/vitest.config.ts
Freymaurer: Personally i like having that log to see if it **really** loads the config files 😅 But i can remove them if we are done moving parts around

tests/vitest.config.ts
Freymaurer: Will add this. I just wanted to get your feedback asap and maybe add CI after the moving parts are mostly done, but it is no big deal

bun.lockb
Freymaurer: Removed bun, as it created issues with gh-action i did not want to find out why 😅

tests/vitest.config.ts
Freymaurer: Done!

tests/vitest.config.ts
MangelMaxime: Let's keep it for now, if it makes too much noise we can remove it later. (I didn't run the PR only review the code)

tests/vitest.config.ts
MangelMaxime: Thank you

paket.dependencies
Freymaurer: > I can make CPM work if you prefer that over Paket

Yes please!

Freymaurer: Done!

PR #659, opened 2025-04-30
Update README.md

kant2002:


Freymaurer:
I think you were a bit slower than #657 😆 Changes are already merged, so i will close the PR

kant2002:
Ooh no, what a loss :)

Issue #652, opened 2025-03-17
React19 and key not found warning


Freymaurer:
Using Feliz with React 19 return a lot of missing key warnings. I investigated this issue and showcase the problem, as well as some solutions

# The Issue

So i took quite some time to get to the bottom of this issue and the culprit is `Interop.reactApi.Children.toArray` using the [react Children api](https://react.dev/reference/react/Children). 

The `Children.toArray` function calls the following, which showcases why we now get these warnings. It still recreates the react element with automatic keys `.0`, `.1`..., but we get the warning.

<details><summary>`toArray` subfunction</summary>
<p>

```js
if (__DEV__) {
// If `child` was an element without a `key`, we need to validate if
// it should have had a `key`, before assigning one to `mappedChild`.
// $FlowFixMe[incompatible-type] Flow incorrectly thinks React.Portal doesn't have a key
if (
  nameSoFar !== '' &&
  child != null &&
  isValidElement(child) &&
  child.key == null
) {
  // We check truthiness of `child._store.validated` instead of being
  // inequal to `1` to provide a bit of backward compatibility for any
  // libraries (like `fbt`) which may be hacking this property.
  if (child._store && !child._store.validated) {
    // Mark this child as having failed validation, but let the actual
    // renderer print the warning later.
    newChild._store.validated = 2;
  }
}
```

</p>
</details> 

# Solution 1: Spreading children

Looking at the documentation for [`React.createElement`](https://react.dev/reference/react/createElement), we can find the following parts:

```js
createElement(type, props, ...children)
```

> optional `...children`: Zero or more child nodes. They can be any React nodes, including React elements, strings, numbers, [portals](https://react.dev/reference/react-dom/createPortal), empty nodes (null, undefined, true, and false), and arrays of React nodes.

> You should only pass children as multiple arguments to createElement if they are all statically known, like createElement('h1', {}, child1, child2, child3). If your children are dynamic, pass the entire array as the third argument: createElement('ul', {}, listItems). This ensures that React will [warn you about missing keys](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key) for any dynamic lists. For static lists this is not necessary because they never reorder

Currently [Feliz handles children by passing them always as props](https://github.com/Zaid-Ajaj/Feliz/blob/master/Feliz/Interop.fs). Before React 19 this was sufficient using the `Children.toArray` method. We can adjust the `createElement` function in Feliz to spread children as third optional parameter:

```fsharp
[...]

Issue #660, opened 2025-05-04
update feliz to jsx syntax, ensure tests pass


MangelMaxime:

Freymaurer:
First Roadblock:

Conditional properties like this are not supported with JSX.create and throws with: `error FABLE: Expecting a static list or array literal (no generator) for JSX props`

```fsharp
let Ele() =
    JSX.create "div" [
        if 1 = 1 then
            "className", "first"
    ]
```

Issue #607, opened 2024-06-03
React 19


laurentpayot:
Hi, React 19 RC has been released a few weeks ago, with some [breaking changes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#breaking-changes).

How will the soon to be released React 19 impact Feliz? Is React 19 compatibility planned?

MangelMaxime:
We should indeed add support for React 19, however I still can't find information on what the expected JavaScript is.

All the documentation, I find are using JSX syntax but for Feliz we need to output result of the JSX compilation.

Currently this is what we are doing:

If in JSX you write

```jsx
const MyButton = () =>  <button />
```

the JavaScript version of that was:

```js
const MyButton = () => /*#__PURE__*/React.createElement("button", null);
```

The equivalent in Feliz is

```fs
[<ReactComponent>]
let MyButtion () =
	Html.button []
```

which compiles to 

```js
import { createElement } from "react";
import React from "react";

export function MyButtion() {
    return createElement("button", {});
}
```

Because of that I don't know yet if adding support for React 19, will be a breaking change or not.

It is also important to note that the code currently generated by Feliz is not even yet up to the latest version of what React does.

Since React 18 (I think), the recommended generated JavaScript is:

```js
import * as JsxRuntime from "react/jsx-runtime";

var MyButton = JsxRuntime.jsx("button", null);
```

This is for example, the reason why when using Fast Refresh with Vite people need to set `jsxRuntime: 'classic'` for it to be working.

Zaid-Ajaj:
Looking at the changes in React 19, nothing strikes me as breaking from Feliz point of view. That said, current Feliz version defines its react dependency (via a transitive reference to `Fable.ReactDom.Types`) as `>= 18.x && < 19` so a breaking change is not out of the question if you happen to upgrade React to v19

MangelMaxime:
I think the easiest way to see if something is broken or not would be for someone to upgrade to React 19 and we will see if React complains or something breaks :)

rasheedaboud:
I updated Feliz to with react 19 bits and the new compiler and it seems to work fine so far...
![image](https://github.com/Zaid-Ajaj/Feliz/assets/49292985/b697e03a-1f0c-4e69-9b95-d8b8d440a59f)

tw0po1nt:
So I ran into an issue with this recently, actually.

[...]

PR #664, opened 2025-05-06
Update Feliz output to JSX

Freymaurer:
Closes: #652, #660, #663, #607


Freymaurer:
Most of the update is done, so i will merge into the epic, to continue working on docs/template project until more work on jsx can be done

HughEdwards1:
Just curious (I've only checked out this project for the last day or two).   
I'm interested in integrating react three fiber with feliz - being that it requires a custom jsx runtime as far as I can tell, I imagine I couldn't lean on the existing interop.reactelement to creating a wrapping for it.

With this work such a thing would be trivial to implement?

HughEdwards1:
Oh really?? I did a vite build and saw it was generating `jsxs` elements imported from the jsx runtime.     
I'm pretty new to this library - 

I might find time to have a play this weekend, but am I right in thinking:   
I could create a file which defines a bunch of Interop.reactElements to define the signature of the jsx elements (eg. `<mesh>`). Then just call them like 
```
[ 
  Html.p "x"
  Three.mesh xxx
]
 ```

That'll compile down into something like
```
React.createElement(p, ...)
React.createElement(mesh, ...)
```

And then vite will somehow figure out the jsx transform there? 

Currently I have the @8 version installed which is what gave me the `jsxs` on vite build. Basically just wondering if another approach would be to put an import statement in of the jsx runtime and then 'import' items from there (and instead of react.createelement using jsx and jsxs depending on children, similar to what I see being done in interop)

Freymaurer:
Vite build will transpile jsx to optimized js functions ddpending on the jsx runtime. For simole react components you can gain a performance boost by using jsx instead of the previous createElement approach. That is one of the reasons why we decided to go for jsx in feliz v3.
But if you just use createElement this is no problem at all for react elements.

You can have a look here for some example bindings, i wrote, for a react library

https://github.com/nfdi4plants/Swate/blob/dbf8439a5fe1b6c77482d69e69273e029fdb4635/src/Components/src/Util/FloatingUI.fs#L244

HughEdwards1:
Super cool mate, thank you for putting that together.   
Keen to have a play and see where I get with it!! :)

Issue #570, opened 2023-03-23
Elmish subscriptions does not work with Feliz.UseElmish


Zaid-Ajaj:
@alfonsogarciacaro :

> @Zaid-Ajaj If I replace Feliz.UseElmish with [Fable.React.UseElmish](https://github.com/fable-compiler/Fable.React.UseElmish/blob/9be9e829a4b1260057bca85e55971790163f194c/src/Fable.React.UseElmish/Fable.React.UseElmish.fs) Elmish subscriptions work. Haven't checked what's the actual difference (maybe because Fable.React.UseElmish doesn't use the useSyncExternalStore polyfill?)
>
> Test project: https://github.com/alfonsogarciacaro/app

_Originally posted by @alfonsogarciacaro in https://github.com/Zaid-Ajaj/Feliz/issues/569#issuecomment-1481192665_

auslavs:
Hey, I was interested in using subscribers in a react component and found myself here.

I forked the example and upgraded it to the latest version of Feliz and it seems to work.

Maybe this issue can be closed?

My fork
https://github.com/[auslavs/elmish-subscriber](https://github.com/auslavs/elmish-subscriber)

PR #683, opened 2025-11-05
Elmish subscriptions does not work with Feliz.UseElmish

auslavs:
This PR is to address #570, where it was reported that Elmish subscriptions did not work with Feliz.UseElmish.

Later versions have already resolved this; However, I was asked if I could contribute tests and documentation.

I have included a test here, which I think runs correctly, but I have never been able to successfully run the whole test collection.
Maybe someone could independently verify, or please give me some pointers for running the tests?

I intend on adding documentation, but I think there should first be some discussion on its usage, at least for my understanding.

Currently, it looks like the only option to add a subscriber is to use the mkProgram function, which would also require passing in some view function, which I'm not sure is relevant in this context and might be confusing.

```fsharp

  [<ReactComponent>]
  let MyComponent () =
    let init () : unit * Cmd<unit> = (), Cmd.none
    let update () () : unit * Cmd<unit> = (), Cmd.none
    let view = fun _ _ -> ()
    let program () = Program.mkProgram init update view
    let state, dispatch = React.useElmish program
    Html.none
```

Is there another way to do this, or is this the preferred way?

If this is the case, then I could prepare some documentation to be added here:
https://fable-hub.github.io/Feliz/api-docs/UseWithElmish

### DIFF
--- a/tests/Tests.fs
+++ b/tests/Tests.fs
@@ -608,6 +608,69 @@ module UseElmish =
         ])
 #endif
 
+#if FABLE_COMPILER_3 || FABLE_COMPILER_4
+module Subscriber =
+    open Elmish
+    open Feliz.UseElmish
+    open Fable.Core.JS
+
+    type Model = {
+        Count: int
+        SubscriberTicks: int
+    }
+
+    type Msg =
+        | Increment
+        | SubscriberTick
+
+    let init initialValue =
+        {
+            Count = initialValue
+            SubscriberTicks = 0
+        },
+        Cmd.none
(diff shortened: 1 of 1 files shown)


auslavs:
Thanks for the tip @Freymaurer ! I'll have a look at v3

auslavs:
I've moved the tests to feliz_3.0, and I'm now able to run them.
However, I am experiencing an issue with my tests that I am currently unable to resolve.
I'll close this PR and open a new one when ready.

PR #643, opened 2025-01-30
Added `menu` HTML element to Html.fs

laurentpayot:
Added the missing `menu` element to Html.fs so it behaves the same as  the `ul` element, as it is a *semantic* alternative: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu

PR #686, opened 2025-11-18
[v3] menu HTML element + popover API props

laurentpayot:
- added the `menu` HTML element from Feliz v2 PR #643
- added the [popover API HTML props](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#html_attributes) with comments from MDN.

Beware I could not test the popover API because for now my project is broken until I replace Feliz.Router with the JS [preact-iso](https://preactjs.com/guide/v11/preact-iso) router.


laurentpayot:
I will try to do this later today or at the beginning of next week. I agree some popover tests would be nice :wink:

Freymaurer:
> even if https://github.com/ionide/ionide-vscode-fsharp/issues/2112 with the test solution 🙄

Strange for me it takes a moment to start up but runs as expected with vs code.

but you can run the tests with?

```
cd .\tests\Feliz\

npm run test:watch
```

PR #707, opened 2026-04-17
Fixed static member inline points (coordinates: string)

ConradRoesch:
`static member inline points (coordinates: string)` was failing to pass the parameter `coordinates` into the function `PropHelper.mkAttr`, returning `e' -> IReactPropery` instead of the desired `IReactPropery`

before:
`    static member inline points (coordinates: string) = PropHelper.mkAttr "points"`

after:
`    static member inline points (coordinates: string) = PropHelper.mkAttr "points" coordinates`


Freymaurer (approved):
Good catch! Please add your contribution to src/Feliz/CHANGELOG.md if you want

Freymaurer (approved):

